PR #65, opened 2019-08-19
Move docs into separate files 📝

bobcatfish:
# Changes

I am about to start adding some configuration to allow Prow to run
Tekton Pipelines (https://github.com/tektoncd/pipeline/issues/922)
and before I do that I wanted to reorganize the docs a bit to make it
easier to add a bit more detail and make the main README a bit less
busy.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #66, opened 2019-08-19
Add a "hello world" Pipeline to be executed via Prow 🌏

bobcatfish:
# Changes

A step toward https://github.com/tektoncd/pipeline/issues/922 !
    
This commit adds configuration and supporting docs to get Prow to execute a simple Pipeline on every pull request to the plumbing repo.
    
I am also working on docs in the Prow repo (for https://github.com/kubernetes/test-infra/issues/13874 and https://github.com/tektoncd/pipeline/issues/1187) to explain how one could put this all together from scratch.
    
One important caveat is that we were previously using the Tekton installation in this cluster to do manual releases, and we were keeping our release configuration up to date with the latest version of Tekton Pipelines. Because of https://github.com/kubernetes/test-infra/issues/13948 the newest version we can use with Prow is v0.3.1 so we need to keep our release configs in sync with this for now ( attn folks doing releases @vdemeester @a-roberts @mnuttall - the next thing I want to do is add release automation via Prow so hopefully this won't impact you for long!)
    
Note I have already manually applied this configuration to the cluster 🙏

p.s. This is branched off https://github.com/tektoncd/plumbing/pull/65 so the diff will get a bit smaller once that one is in!

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -1,0 +1,96 @@
+# Prow
+
+`tektoncd` uses
+[`Prow`](https://github.com/kubernetes/test-infra/tree/master/prow)
+for CI automation.
+
+- Prow runs in [the tektoncd GCP project](../gcp.md)
+- [Ingress is configured to `prow.tekton.dev`](#ingress)
+- Prow results are displayed via [gubernator](../gubernator/README.md)
+- [Instructions for updating Prow](#updating-prow-itself)
+- [Instructions for updating Prow configuration](#updating-prow-configuration)
+
+_[Prow docs](https://github.com/kubernetes/test-infra/tree/master/prow)._
+_[See the community docs](../CONTRIBUTING.md#pull-request-process) for more on
+Prow and the PR process._
+
+### Secrets
+
+- Prow uses the service account
+  `prow-account@tekton-releases.iam.gserviceaccount.com`
+- Secrets for this account are configured in [Prow's config.yaml](config.yaml) via
+  `gcs_credentials_secret: "test-account"`
+
+## Ingress
+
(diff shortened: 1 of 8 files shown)


bobcatfish:
image

image

🎉 🎉 🎉

bobcatfish:
btw it's hard to find the logs from this at the moment, but I'm working on a solution with @sbwsg which would at least make it so that maintainers could easily jump to the logs in stackdriver (should be available automatically - I think I need to upgrade the prow cluster for that tho :pray: )

prow/README.md
```diff
@@ -0,0 +1,121 @@
...
+  The intention seems to be that `test-pods` be used to run the pods themselves, but we
+  don't currently have that configured in our [config.yaml](config.yaml).
+
+#### Tekton Pipelines with Prow
+
+[Tekton Pipelines](https://github.com/tektoncd/pipelines) is also installed in the `prow`
+cluster so that Prow can trigger the execution of
+[`PipelineRuns`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md).
+
+[Since Prow only works with select versions of Tekton Pipelines](https://github.com/kubernetes/test-infra/issues/13948)
```
goern, line 86: Is this statement still true?

rlandesman:
How would one inject either static or runtime variables into Tasks using Prow? When running manually with a PipelineRun I had them defined there but now that Prow is triggering the Pipeline I'm unsure of where to place them.

I'm an intern so a bit lost in all of this; thanks in advance for the help!

PR #68, opened 2019-08-22
Add docs on how to create a new Prow cluster ✨

bobcatfish:
# Changes

In the context of getting a Tekton pipeline running on Prow
(https://github.com/tektoncd/pipeline/issues/922), I have been working
with @sbwsg on how to make logs visible from the running Pipeline
(as of #66 we have a simple hello world type pipeline running against
the plumbing repo itself). Prow normally uploads logs to GCS itself via
a sidecar, but Prow + Tekton doesn't do this.

@sbswg discovered that GKE clusters automatically export logs to
Stackdriver and propogate all of the pod labels as well, which means we
should be able to look in Stackdriver and filter logs by PipelineRun
names or Pipeline names or whatever we'd like. This could be an okay
interim solution that would allow us to migrate some work from ProwJobs
to Pipelines, e.g. we can add Release Pipelines and give other OWNERS
access to the Stackdriver logs.

However our Prow cluster was too old to have this feature! I was under
the mistaken impression that I had to make a whole new cluster, so I
started doing that (and writing down how to do it), but then @dlorenc
brought to my attention that instead we can just update the master +
nodes in GKE. So I updated the master node and now we have the feature!

In the future we should be able to keep updating our cluster like this,
but I thought it would still be useful to know how to create the cluster
from scratch so I wrote down what I discovered.

I didn't get as far as updating the DNS entry with the ingress IP which
is why that part is the most sparse but hopefully better than nothing!

I also updated our Boskos configuration with some missing Resources that
are in the kubernetes/test-infra config, and deployed to our cluster, 
but now in our committed config.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -7,6 +7,7 @@ for CI automation.
 - Prow runs in [the tektoncd GCP project](../gcp.md)
 - [Ingress is configured to `prow.tekton.dev`](#ingress)
 - Prow results are displayed via [gubernator](../gubernator/README.md)
+- [Instructions for creating the Prow cluster](#creating-the-prow-cluster)
 - [Instructions for updating Prow](#updating-prow-itself) and [Prow's Tekton Pipelines instance](#tekton-pipelines-with-prow)
 - [Instructions for updating Prow configuration](#updating-prow-configuration)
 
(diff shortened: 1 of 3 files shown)


prow/README.md
```diff
@@ -7,31 +7,132 @@ for CI automation.
...
+ --machine-type=n1-standard-4 \
+ --image-type=cos \
+ --num-nodes=8 \
+ --cluster-version=latest
+```
+
+### Creating the secrets
+
+In order to operate, Prow needs the following secrets, which are referred
+to by the following names in out config:
```
sbwsg: `out` -> `our` ?

prow/README.md
```diff
@@ -7,31 +7,132 @@ for CI automation.
...
+
+- `GCP` secret: `test-account` is a token for the service account
+  `prow-account@tekton-releases.iam.gserviceaccount.com`. This account can
+   interact with GCP resources such as uploading Prow results to GCS
+   (which is done directly from the containers started by Prow, configured in [config.yaml](config.yaml)) and
+   [interacting with boskos clusters](../boskos/README.md).
+- `Github` secrets: `hmac-token` for authtenticating GitHub and `oauth-token` which is a
```
sbwsg: `authtenticating` -> `authenticating`

prow/README.md
```diff
@@ -68,7 +169,9 @@ These values have been removed from the original
 [starter.yaml](https://github.com/kubernetes/test-infra/blob/master/prow/cluster/starter.yaml):
 
 - The `ConfigMap` values `plugins` and `config` because they are generated from
-  [config.yaml](config.yaml) and [plugins.yaml](plugins.yaml)
+  [config.yaml](config.yaml) and [plugin
```
sbwsg: Did the line get cut here intentionally?  Looks like `# Deploy boskos` was accidentally pasted into the plugin link markdown?

prow/README.md
bobcatfish: wow i can barely see the difference XD :eyeglasses:

prow/README.md
bobcatfish: whoooooops...

PR #72, opened 2019-09-17
Remove implicit self approve

afrittoli:
The implicit self approve means that every PR submitted by an owner
may be inadvertently approved and merged by any member of the
tektoncd org with an /lgtm, where it should generally speaking be
reviewed and approved first by another member of the OWNERS team.
I think it's best to err on the safe side :)

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -15,7 +15,7 @@
 approve:
 - repos:
   - tektoncd
-  implicit_self_approve: true
+  implicit_self_approve: false
   review_acts_as_approve: true
 
 plugins:


bobcatfish:
Sounded like @abayer had strong feelings!

bobcatfish:
p.s. alternatively, i wonder if the real problem is being more restrictive in the 'lgtm' powers??

bobcatfish:
I could go either way on this one! Personally I think in most cases its fine for an OWNER to get approval and a random other org member lgtm - in my mind the real issue is around API changes (https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#approving-api-changes) and it would be nice to have something to make sure that API changes get the appropriate level of review. So my 2 cents are that removing the implicit self-approval doesn't really do much to address our real issues, but if folks want to go ahead with this then I'm okay with it!

bobcatfish:
Latest in my stream of consciousness thoughts: maybe we can just try this and see how it goes?

/lgtm

(hee hee hee)

afrittoli:
Heh, yeah, this would not be written in stone, but as we move towards 1.0 I think having two reviewers is probably not a bad thing, more likely to catch accidental breaking changes.

vdemeester:
/hold cancel
:tada:

PR #73, opened 2019-09-24
Add a pipeline to deploy and test a release of a Tekton project

afrittoli:
Add a Task that gets a release of a Tekton project from the release
bucket, and installs it in a target cluster.

# Changes

# Submitter Checklist

review them:

for more details._


afrittoli:
Added "hold" to allow for a couple of reviews. This is not used anywhere yet but I'm building a pipeline to test releases bit by bit.

afrittoli:
Related to #75

scripts/logs/README.md
```diff
@@ -0,0 +1,48 @@
+# How to follow log outputs
```
vdemeester: Shouldn't we just point to `tkn` instead ?

tekton/README
```diff
@@ -0,0 +1,213 @@
+# Tekton Resources for CI/CD
```
vdemeester: The file should be `README.md` :stuck_out_tongue:

afrittoli: heh, nice catch, thanks

scripts/logs/README.md
afrittoli: this is copied straight from pipeline - we might do something different eventually - but I wanted to have a smooth migration, by using the existing scripts mostly wrapped into tasks

afrittoli:
I think this is in a reasonable state to go into the repo now, and then iterate from here.

tekton/README
afrittoli: Done

scripts/logs/README.md
afrittoli: I've seen your PR about this, once it's done on pipelines I can port it to plumbing if that's ok.

PR #76, opened 2019-09-30
Add an app to expose PipelneRun logs publicly

sbwsg:
# Changes

Prior to this commit we didn't have a way to show public logs for PipelineRuns started by a Prow job.

This commit adds a simple go app to expose logs from PipelineRuns kicked off by Prow via the Stackdriver API.

The app exposes an http endpoint that receives a Prow Build ID as a query param and then looks up the logs for that build ID. This requires an updated version of Prow to expose the BuildID to the job_url_template. (added here: https://github.com/kubernetes/test-infra/pull/14322)

# Submitter Checklist

review them:


sbwsg:
Screenshot of the results in a browser:

sbwsg:
> If we label our pipelineruns with a BuildID, would that be enough, or is there any other requirement for prow integration?

Right, it would need a label with name `prow.k8s.io/build-id"` and value like `1234567`.

afrittoli:
@sbwsg  Could you add as a follow-up a CI job to run the tests?
Is there a way with prow to run specific tests only if the change affects specific folders?

PR #100, opened 2019-10-29
Configure stackdriver access for pipelinerun-logs app

sbwsg:
# Changes
The pipelinerun-logs app needs access to stackdriver APIs in order to show log entries for pipelineruns. GCP supports a feature called workload identities that enables a k8s serviceaccount to refer to an IAM serviceaccount and use its credentials when accessing google cloud apis. This means we don't need to download JSON keys and inject them as secrets for apps running in the cluster to utilize cloud apis.

I've created a service account in GCP that has access to stackdriver, I've created a corollary service account in k8s (in the default namespace) that references the GCP service account, and I've created a node pool that has the workload identity metadata service running. The pipelinerun-logs deployment has been updated to schedule its pods on the new nodes and after manual testing the app seems to now have access to stackdriver APIs.

# Submitter Checklist

review them:

### DIFF
--- a/pipelinerun-logs/config/deployment.yaml
+++ b/pipelinerun-logs/config/deployment.yaml
@@ -14,6 +14,9 @@ spec:
       labels:
         app: pipelinerun-logs
     spec:
+      # this service account has access to the stackdriver API
+      # via workload identity
+      serviceAccountName: pipelinerun-logs-viewer
       containers:
       - name: pipelinerun-logs
         image: github.com/tektoncd/plumbing/pipelinerun-logs/cmd/http
@@ -30,3 +33,6 @@ spec:
         - "9999"
         ports:
         - containerPort: 9999
+      nodeSelector:
+        # Schedule this deployment onto nodes with workload identity enabled
+        iam.gke.io/gke-metadata-server-enabled: true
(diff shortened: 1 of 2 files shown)


sbwsg:
These were the steps I followed to enable workload identities on a new node pool: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

pipelinerun-logs/config/deployment.yaml
```diff
@@ -30,3 +33,6 @@ spec:
         - "9999"
         ports:
         - containerPort: 9999
+      nodeSelector:
```
dlorenc: Can we just enable this on all nodes?

sbwsg: Yes it's possible.  Initially I didn't want to interfere with any existing workloads running on the non-workload-identity nodes so I created a separate node pool just for the feature.

I'm not clear on what the risks are for workloads on migrated nodes - i.e. what is the default behaviour if a service account wasn't originally included in a Deployment that is then migrated to a workload identity node?

vdemeester:
@afrittoli @sbwsg it is already deployed on the `dogfooding` cluster right ?

sbwsg:
yes that's right

afrittoli:
Thanks for this!

PR #121, opened 2019-11-22
Add license tools for repositories without vendor directories.

wlynch:
# Changes

Needed for upcoming work for
https://github.com/tektoncd/pipeline/issues/1538.

This change makes 2 changes to support Go module enabled repos
without vendor directories.

1) Use go-licenses instead of dep-collector for third_party license
   notices. dep-collector makes heavy assumptions about the existence of
   the vendor directory. go-licenses is a module aware alternative.
2) Ignore the `+build tools` flag when building/testing projects. This
   is to allow tool dependencies to be specified with Go modules following
   the recommendation of
   https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module.

# Submitter Checklist

review them:

for more details._


scripts/presubmit-tests.sh
```diff
@@ -167,7 +167,7 @@ function default_build_test_runner() {
   go build -v ./... || failed=1
   # Get all build tags in go code (ignore /vendor)
   local tags="$(grep -r '// +build' . \
-      | grep -v '^./vendor/' | cut -f3 -d' ' | sort | uniq | tr '\n' ' ')"
+      | grep -v '^./vendor/' | grep -v '^./third_party/' | grep -v "+build tools" | cut -f3 -d' ' | sort | uniq | tr '\n' ' ')"
```
vdemeester: @wlynch why the `grep -v "+build tools"` ?

wlynch: This is to exclude packages containing binary dependencies, as specified by https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module.

Added a comment to document.

vdemeester: Ah nice, thanks :+1:

PR #130, opened 2019-11-28
Add a image with `hub` and a nightly build job

afrittoli:
Hub is a CLI interface to GitHub. This image will be used to run release automation scripts.
Add a cronjob that builds the image nightly to tag "latest".

PR #129, opened 2019-11-28
 Install a pinned stable version of golangci-lint  instead of master

chmouel:
# Changes

 Install golangci-lint from a pinned version

Let's install golangci-lint from a stable pinned version instead of master which
can brings weird CI breakage.

Closes #125

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [?] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #132, opened 2019-11-28
Use tag latest for test-runner instead of a digest 😅

vdemeester:
# Changes

This should allow us to update the test-runner without having to bump
the prow config each time.

Related to https://github.com/tektoncd/plumbing/pull/130#discussion_r351765456 and https://github.com/tektoncd/plumbing/pull/129

/cc @chmouel @afrittoli @wlynch 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -84,7 +84,7 @@ presubmits:
     trigger: "(?m)^/test (all|pull-tekton-cli-build-tests),?(\\s+|$)"
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


chmouel:
/lgtm

nice one, thanks!

vdemeester:
> The only concern I have with this is that the testrunnner image pulls in the latest version of some tools, and since this is built nightly, it means we will continuously be updating the image, which sooner or later will cause some CI job to break.
Right, this is why we should try to use released version when we can — that way, an update will need to be explicit but we won't need to do 2 PRs for the projects to benefit from it.

PR #133, opened 2019-11-28
Fixes after PR#131

afrittoli:
PR#131 made the task to build images more discoverable but it also
introduced a few issues. Fixing those. This version is now running
succesfully on the dogfooding cluster.

### DIFF
--- a/tekton/config/nightly-image-build-cron-base/trigger-image-build.yaml
+++ b/tekton/config/nightly-image-build-cron-base/trigger-image-build.yaml
@@ -42,28 +42,32 @@ spec:
           containers:
           - name: trigger
             image: curlimages/curl
-            script: |
-              set -ex
+            command:
+            - /bin/sh
+            args:
+              - -ce
+              - |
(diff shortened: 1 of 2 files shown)


tekton/resources/image-build-trigger.yaml
```diff
@@ -47,13 +47,13 @@ spec:
   - name: contextPath
     description: The path to the context within 'gitRepository'
   - name: registry
-    description: The container registry: *registry*/namespace/name:tag
```
afrittoli: yaml not happy about colons here...

tekton/config/nightly-image-build-cron-base/trigger-image-build.yaml
```diff
@@ -42,28 +42,32 @@ spec:
           containers:
           - name: trigger
             image: curlimages/curl
-            script: |
```
afrittoli: heh, not a task, just plain container :P

PR #135, opened 2019-12-02
Add the letsencrypt cert-manager configs.

dlorenc:
# Changes

We had an old one that stopped working due to Let's Encrypt changes.
This fixes that by upgrading cert-manager and our configs.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/clusterissuer.yaml
+++ b/prow/clusterissuer.yaml
@@ -1,0 +1,14 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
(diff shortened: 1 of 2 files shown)


prow/clusterissuer.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cert-manager.io/v1alpha2
```
afrittoli: NIT: I wonder if this should live under the `prow/` folder since it's something that we can use to obtain certificate for all ingresses in the cluster. But no reason to hold on this PR for that.

PR #149, opened 2019-12-10
Refactor pipeline nightly-release into an kustomize overlay.

wlynch:
# Changes

This is the first step to make it easy to reuse common components for
nightly releases for other project (e.g. triggers - https://github.com/tektoncd/triggers/issues/237).

 Not all resources followed the same name scheme (e.g. nightly-release vs release-nightly, different prefixes, etc), so this is *not* a no-op in terms of resources being defined. Generated yaml: https://gist.github.com/wlynch/ea08255baf443824630801ef31e13f45

# Submitter Checklist

review them:

for more details._


tekton/resources/nightly-release/base/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
```
afrittoli: Is the idea to have one file per resource type?
We might need more service accounts in the future. Currently we one service account to run the release event listener and then to run the release, but we may want to split that our in future on one SA for the listener that has no access to secrets, one a different one to run the release that does have access to secrets.

tekton/resources/nightly-release/overlays/pipeline/template.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: versionTag
+        value: $(params.versionTag)
+      resources:
+      - name: source-repo
+        resourceRef:
+          name: git-source-$(uid)
+      - name: bucket
+        resourceRef:
+          name: tekton-bucket-nightly-$(uid)
+      - name: builtBaseImage
```
afrittoli: We don't really benefit from using the image resource here - I wonder if we should just switch to an array of targets. Or perhaps we could support a new input type "list of resources" :)

tekton/resources/nightly-release/base/serviceaccount.yaml
wlynch: Right now it doesn't make a big difference, but splitting each resource into different files seems like it would make it easier to support base reuse across different kustomize configs in the future. e.g. have one multiple base configs that reuse the same EventListener, Template, but doesn't create its own SA.

I don't think how this is currently laid out limits any changes we would want to make for service account usage later.

tekton/resources/nightly-release/overlays/pipeline/template.yaml
wlynch: This is just copying the existing behavior: https://github.com/tektoncd/plumbing/blob/e34ec9c0a73900e953b726d93343c61e462bb4eb/tekton/resources/pipeline-nightly-release/pipeline-nightly-build-trigger.yaml#L96-L135

We need to at the moment because of the pipeline Pipeline: https://github.com/tektoncd/pipeline/blob/5af2878a30cf7a65e97bfe052bd00dd2a7133c09/tekton/release-pipeline-nightly.yaml#L13-L37.

afrittoli: Yes, absolutely, this was just a general comment on things we might do in future, not for this patch for sure :)

tekton/resources/nightly-release/base/serviceaccount.yaml
afrittoli: ok

afrittoli:
@vdemeester ?
/lgtm

tekton/resources/nightly-release/overlays/pipeline/template.yaml
vdemeester: I tend to agree with you @afrittoli, but yeah, this is something to think about for later :angel:

afrittoli:
@wlynch can you add docs on how to apply this in the next patch?

PR #170, opened 2019-12-20
Make labels a bit easier for build cops 👮‍♀️

bobcatfish:
# Changes

I was build cop today and I noticed in our "build cop runbook" we
talked about the purposes of a few labels, but definitely not all of
them. Since this is the authoritative source on what labels we have, I
think it makes sense to make this our documentation as well, and also to
clean out labels we don't need (many are copied from knative which
copied from kubernetes)

Removed some labels:
- triage/duplicate: Can use "Duplicate of" instead
  (https://help.github.com/en/github/managing-your-work-on-github/about-duplicate-issues-and-pull-requests)
- triage/needs-information: Could use zenhub "waiting for response"
  and/or "kind/question" instead
- triage/not-reproducible, triage/support, triage/unresolved - not
  really sure what these would mean for us
- kind/api-change, "api-review": Can use "area/api" instead
- area/release, area/dogfooding, area/test-infra - We have a separate
  plumbing repo where we can put these issues instead
- kind/failing-test, kind/flake - Seems oddly specific. Instead IMO this is kind of
  bug, and in fact something we should drop everything to fix.
- lifecycle/* - We do not use this plugin
- needs-kind, needs-sig, release-note*, shrug - We don't use these

Remove labels from repos that aren't used:
- cli: user-study
- community: area/cdf-summit
- pipeline: maybe-next-milestone (can use priority instead), user-study,
  also old silly labels like "meaty-juicy-coding-work", okr (not using
  anymore)

Removed labels from repos that were also global

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -61,12 +61,24 @@ default:
       prowPlugin: label
       addedBy: anyone
     
-    # Kinds and areas
+    # Areas
     - color: 0052cc
       description: Indicates an issue or PR that deals with the API.
       name: area/api
       target: issues
       addedBy: label
+    - color: 7057ff
+      description: Indicates an issue on release (process, tasks).
+      name: area/release
+      target: issues
+      addedBy: label
+    - color: 10677f
+      description: Issues or PRs related to the plumbing infrastructure but needing attention in a non-plumbing repo.
+      name: area/plumbing
+      target: both
+      addedBy: label
+    
+    # Kinds
     - color: e11d21
       description: Categorizes issue or PR as related to a bug.
       name: kind/bug
@@ -75,6 +87,9 @@ default:
       target: both
       prowPlugin: label
       addedBy: anyone
+    - color: f7c6c7
+      description: Categorizes issue or PR as related to a flakey test. Please escalate this and treat it as an emergency.
+      name: kind/flake
     - color: c7def8
       description: Categorizes issue or PR as related to cleaning up code, process, or technical debt.
       name: kind/cleanup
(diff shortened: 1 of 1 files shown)


vdemeester:
@bobcatfish few comments :angel: :stuck_out_tongue_closed_eyes: 

>     * triage/duplicate: Can use "Duplicate of" instead
>       (https://help.github.com/en/github/managing-your-work-on-github/about-duplicate-issues-and-pull-requests)

The useful thing about the flag is that we can filter on it. I don't see a way to filter duplicates (using `Duplicates of`.

>     * triage/needs-information: Could use zenhub "waiting for response"
>       and/or "kind/question" instead

1. Not everybody has zenub :stuck_out_tongue: 
2. `kind/question` can be waiting for response from the owner of the code, this label is more about waiting for response from the creator of the issue

>     * area/release, area/dogfooding, area/test-infra - We have a separate
>       plumbing repo where we can put these issues instead

I am not sure on removing those. We do have `plumbing` *but* each project manage their release, and a little bit of their *test-infra* (images to use, …) ; and they will have more control in the future once using tekton :angel:. For `dogfooding` I do agree it make sense almost only on `plumbing` :stuck_out_tongue_closed_eyes: 

>     * kind/failing-test, kind/flake - Seems oddly specific. Instead IMO this is kind of
>       bug, and in fact something we should drop everything to fix.

I like having `kind/testing` and `kind/flake` on an issue (`kind/failing-test` can go :angel:), because some `kind/testing` could be enhancement, or something else :angel:.
 
>     * lifecycle/* - We do not use this plugin

+:100: 

>     * needs-kind, needs-sig, release-note*, shrug - We don't use these

Agree except I do like `shrug` :joy: 
/shrug
 
Otherwise :+1:

bobcatfish:
> The useful thing about the flag is that we can filter on it. I don't see a way to filter duplicates (using Duplicates of.

Ah okay, interesting point - why do you need to filter on it?

> Not everybody has zenub :stuck_out_tongue:

Who doesn't have it? Can we add them?

> kind/question can be waiting for response from the owner of the code, this label is more about waiting for response from the creator of the issue

Okay - seems to me like the action is the same in either case tho, right? Like in both cases we want to keep an eye on them and check back to see if folks have responded?

Hoping to discuss zenhub + these things more in the productivity working group :D Do you want me to add triage/needs-information back for now?

> I am not sure on removing those. We do have plumbing but each project manage their release, and a little bit of their test-infra (images to use, …) ; and they will have more control in the future once using tekton angel.

I see what you mean! Are we using them right now? We could add them back as we start using them if not.

> I like having kind/testing and kind/flake

To me it comes down to whether we'd use these or not: if we have too many labels, then not everyone triaging will use all of them and we won't be able to rely on them. Also IMO kind/flake should be like: drop everything emergency so I hope we won't be filtering on these for long.

@vdemeester sounds like you'd like me to add back:
* duplicates
* needs-information
* flake
* some of the test-infra ones

I'm just not sure that they're useful unless everyone is using them

bobcatfish:
Oh and /shrug I guess XD

bobcatfish:
Okay I think I've re-added the labels you mentioned, lemme know what you think @vdemeester !

PR #185, opened 2020-01-17
Migrate resources to work with triggers v0.2.0

### LINKED ISSUE #183: Upgrade triggers to release v0.2.0
This issue is meant to track the work on migrating to triggers to v0.2.0 in the dogfooding cluster.
https://github.com/tektoncd/triggers/releases/tag/v0.2.0

1. Upgrade the running service
    * it could just be a kubectl apply <release-file>, since we don't have any special configuration in place
    * it would be good to try and use the "deploy tekton service" pipeline instead, and make sure we have overlays in place if needed. That will make it easier to switch to CD of the service in future
1. Upgrade the resources from the plumbing repo:
    * Upgrade bindings to use JSONPath syntax
    * Upgrade event listener to use bindings (as opposed to binding)
    * Upgrade event listener to use interceptors (as opposed to interceptor)
    * Update deployed resources
1. Upgrade the Mario bot
    * Update Mario to exploit the inbuilt github interceptor and CEL filters
    * Update the deployed Mario bot
1. Upgrade the Prow config CD mechanism
    * The solution runs on dogfooding but the PR is not merged yet
    * Update the PR
    * Redeploy resources
1. Test periodic jobs, mario and prow config cd

afrittoli:
Migrate resources to triggers v0.2.0:
    - eventlistener.binding -> bindings
    - eventlistener.interceptor -> interceptors
    - gjson to JSONPath in bindings
      - select an element from an array

Fixes #183

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -12,3 +12,5 @@ Resources are organised in folders:
 - The [resources](resources/README.md) folder contains Tekton resources used for
   various automation tasks: building container images, doing releases,
   maintaining the GitHub org and more.
+- The [cd](cd/README.md) folder contains kustomize overlays, used to deploy the
+  various Tekton projects to the infra clusters.
--- a/tekton/cd/README.md
+++ b/tekton/cd/README.md
@@ -1,0 +1,31 @@
+# (Continuous) Deployment of Tekton Services
+
+This folder includes overlays used to maintain the configuration of Tekton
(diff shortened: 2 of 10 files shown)


abayer:
My apologies for the comment notification spam - I was trying to test #1185 on itself and, well, I put `185` in as the PR number in my test instead and it kept commenting here while I was baffled as to why it was claiming to succeed but never posting a comment over there. 🤦

Issue #186, opened 2020-01-18
Boskos seems to be wedged


bobcatfish:
# Expected Behavior

Boskos should clean up projects once they are done being use and make them available for future use.

# Actual Behavior

https://github.com/tektoncd/pipeline/pull/1541 and https://github.com/tektoncd/pipeline/pull/1888 both have consistently failing integration tests with an error like:

```
I0117 21:16:30.477] 2020/01/17 21:16:30 main.go:734: provider gke, will acquire project type gke-project from boskos
I0117 21:21:30.475] 2020/01/17 21:21:30 main.go:316: Something went wrong: failed to prepare test environment: --provider=gke boskos failed to acquire project: resources not found
```

In #29 and other times in the past we have responded to this error by provisioning more projects for boskos.

This time though it's definitely not the case that all the projects are in use:

* https://pantheon.corp.google.com/home/activity?project=tekton-prow-9 <-- had current activity
* https://pantheon.corp.google.com/home/activity?project=tekton-prow-10 <-- hasnt had an activity since the 14th

When I look at the logs from the boskos Janitor I see this kind of error:

```
 msg: "failed to clean up project tekton-prow-10, error info: Activated service account credentials for: [prow-account@tekton-releases.iam.gserviceaccount.com]
ERROR: (gcloud.compute.instances.list) Some requests did not succeed:
 - Invalid value for field 'zone': 'asia-northeast3-a'. Unknown zone.
 - Invalid value for field 'zone': 'asia-northeast3-b'. Unknown zone.
 - Invalid value for field 'zone': 'asia-northeast3-c'. Unknown zone.

Fail to list resource 'instances' from project 'tekton-prow-10'
ERROR: (gcloud.compute.disks.delete) unrecognized arguments: --global 

To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS
Error try to delete resources: CalledProcessError()
ERROR: (gcloud.container.clusters.list) ResponseError: code=404, message=Not Found.
[=== Start Janitor on project 'tekton-prow-10' ===]
[=== Activating service_account /etc/test-account/service-account.json ===]
[=== Finish Janitor on project 'tekton-prow-10' with status 1 ===]
```

I _think_ the gcloud error might be a red herring, maybe a state that boskos gets into after some other kind of error first.

CPU and memory usage for both boskos + the boskos janitor started going up a few hours ago but its hard to say if that is causing the problem or if the problem is causing it:

[...]

PR #187, opened 2020-01-18
Flailing around trying to get boskos to work

bobcatfish:
Trying to address https://github.com/tektoncd/plumbing/issues/186 but it
does nothing

Just opened the PR to reflect the change I've made manually in the cluster

### DIFF
--- a/boskos/boskos-config.yaml
+++ b/boskos/boskos-config.yaml
@@ -17,6 +17,7 @@ data:
       - tekton-prow-11
       - tekton-prow-12
       - tekton-prow-13
+      - tekton-prow-14
       state: dirty
       type: gke-project
 kind: ConfigMap


boskos/boskos-config.yaml
```diff
@@ -17,6 +17,7 @@ data:
       - tekton-prow-11
       - tekton-prow-12
       - tekton-prow-13
+      - tekton-prow-14
```
afrittoli: I dropped this from the cluster config as boskos was complaining about it:
```
ERROR: (gcloud.compute.sole-tenancy.node-templates.list) HTTPError 403: Access Not Configured. Compute Engine API has not been used in project [censored] before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=[censored] then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.\nFail to list resource 'sole-tenancy' from project 'tekton-prow-14'
```

PR #203, opened 2020-01-28
Add tasks/buildpacks label

natalieparellano:
Resolves https://github.com/buildpacks/tekton-integration/issues/3

This label will help funnel issues to the buildpacks team so that they do not get lost.

cc @jromero @vdemeester

# Changes

Add tasks/buildpacks label

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/label_sync/labels.yaml
+++ b/label_sync/labels.yaml
@@ -256,6 +256,13 @@ default:
       prowPlugin: size
       addedBy: prow
 repos:
+  tektoncd/catalog:
+    labels:
+      - color: DF0A6B
+        description: This issue needs attention from the buildpacks team
+        name: tasks/buildpacks
+        target: issues
+        addedBy: humans
   tektoncd/dashboard:
     labels:
       - color: 160a72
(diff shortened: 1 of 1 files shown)


label_sync/labels.yaml
```diff
@@ -256,6 +256,13 @@ default:
       prowPlugin: size
       addedBy: prow
 repos:
+  tektoncd/catalog:
+    labels:
+      - color: DF0A6B
+        description: This issue needs attention from the buildpacks team
+        name: tasks/buildpacks
```
afrittoli: NIT: I think area/buildpacks would be more consistent with the other labels we use.  The kind/* labels can be used to distinguish bugs from features etc.

jromero: I think we chose `tasks/` so that it's not conflated with other labels that are not related to the catalog "entries". The idea would be that `tasks/*` is associated with and the entry `tekton/catalog/*`.

We wouldn't be opposed to changing it. That was just our train of thought.

natalieparellano: I went ahead and changed the label to area/buildpacks

natalieparellano:
@googlebot I signed it!

natalieparellano:
Apologies for the delay and multiple force pushes - the commits should now be signed properly and hopefully the CLA is satisfied.

natalieparellano:
@googlebot I signed it!

PR #218, opened 2020-02-06
Deployment overlay for the dashboard in dogfooding

### LINKED ISSUE #182: Setup controlled access to Tekton Dashboard for the dogfooding cluster
# Expected Behavior
Project maintainers should be able to use the Tekton Dashboard for the dogfooding cluster, with no need to have direct access to the k8s cluster itself.

# Actual Behavior
The only way to access the cluster today is via port-forwarding.

# Additional Info

1. We'll need a dedicated subdomain of tekton.dev
2. We need an auth system, it could be a web-server with HTTP basic auth or [IAP on GKE](https://cloud.google.com/iap/docs/enabling-kubernetes-howto)
3. We need automation and documentation for the whole setup

Eventually it might be nice to get authentication integrated in the tekton dashboard itself.

afrittoli:
# Changes

Add an overlay for the dashboard in the dogfooding cluster.
The overlay includes:
- an ingress resource
- setting the service to "nodePort", which is required for
  the ingress to work with GKE.

This patch includes also a ClusterIssuer that can be used for
any ingress in the cluster. Using the cluster issuer requires
cert-manager to be installed first:

https://cert-manager.io/docs/installation/kubernetes/#installing-with-regular-manifests

GKE load balancer setup the health check to point to the
readiness probe defined in the pods associated to the service.
The probe is expected to return 200, while in the latest
dashboard it returns 204 No Content. This patch fixes the issue:
https://github.com/tektoncd/dashboard/pull/1017

Partially fixes #182

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cd/dashboard/base/kustomization.yaml
+++ b/tekton/cd/dashboard/base/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+- release.yaml
--- a/tekton/cd/dashboard/base/release.yaml
+++ b/tekton/cd/dashboard/base/release.yaml
@@ -1,0 +1,305 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: extensions.dashboard.tekton.dev
+spec:
+  group: dashboard.tekton.dev
+  names:
+    categories:
+    - tekton
+    - tekton-dashboard
+    kind: Extension
+    plural: extensions
+  scope: Namespaced
+  subresources:
+    status: {}
+  version: v1alpha1
+---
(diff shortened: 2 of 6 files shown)


dibyom:
Curious about the ingress -- is this the GKE ingress or a nginx ingress that we install on cluster? For the latter I think clusterIP services should also work.

/lgtm

dibyom:
nvm -- found the answer at https://github.com/tektoncd/dashboard/pull/1017 :smile: 

"The GKE Ingress uses an external loadbalancer with healthchecks that expect 200 OK back instead of 204 no content"

Issue #193, opened 2020-01-21
Replace Boskos StatefulSet with Deployment


bobcatfish:
# Expected Behavior

If you make changes to the boskos configuration and apply it, you should be able to see the changes in the resulting pod.

# Actual Behavior

The boskos StatefulSet has update strategy "ondelete" so you have to delete the pod before it'll get the update.

# Steps to Reproduce the Problem

1. Change the image in boskos/boskos.yaml
2. Apply the config
3. Watch nothing happen

# Additional Info

The k8s prow folks seem to have gone through this same cycle https://github.com/kubernetes/test-infra/issues/11956 and are now using a Deployment (and they got rid of the PVC as well) https://github.com/kubernetes/test-infra/pull/13594/files

afrittoli:
@bobcatfish is this something we still need to fix?

bobcatfish:
@afrittoli yeah, esp. since it looks like we're for sure going to be using boskos in the future. @tomgeorge even has the solution in #225, we just need to verify that it works. I can add this to my todo list!

bobcatfish:
feels like something we should do but if it's not important enough then maybe it's not worth it

PR #225, opened 2020-02-09
Change Boskos StatefulSet to Deployment.

tomgeorge:
# Changes

This change deploys Boskos as a Deployment rather than a StatefulSet, to enable quicker feedback in the deployment process, and removing the need to delete the Boskos pods before new pods are started.

This addresses #193

# Submitter Checklist

review them:

- [] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/boskos/boskos.yaml
+++ b/boskos/boskos.yaml
@@ -83,26 +83,6 @@ kind: Namespace
 metadata:
   name: test-pods
 ---
-apiVersion: v1
-kind: PersistentVolume
-metadata:
-  labels:
-    app: boskos
-  name: boskos-storage
-  namespace: test-pods
-spec:
-  claimRef:
-    name: boskos-volume-boskos-0
-    namespace: test-pods
-  capacity:
-    storage: 1Gi
-  accessModes:
-    - ReadWriteOnce
-  persistentVolumeReclaimPolicy: Retain
-  gcePersistentDisk:
-    pdName: boskos-storage
-    fsType: ext4
----
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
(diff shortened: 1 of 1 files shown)


tomgeorge:
Sorry for the delay. I wasn't super sure either, I had faith that test-infra did the right thing :sweat_smile: 

Looking into it a little more, I think Boskos keeps a list of resources that are stored in the PVC.  See https://github.com/kubernetes/test-infra/blob/2ec0aac72450a8617baf78066a08bb7e943289ee/boskos/common/common.go#L103 for the definition of the `common.Resource` struct.  

The storage plugin for boskos adds/removes those `Resources` to the PVC. 

Boskos added the ability to [store that user data as CRDs](https://github.com/kubernetes/test-infra/tree/master/boskos/crds), so I think going forward they have migrated to that.  The Resource information (Name, Type, Last Updated, etc) are stored in the API server.

tomgeorge:
To clarify, it's storing the current state of Boskos and the resources under management

sbwsg:
/hold

/approve

@tomgeorge @afrittoli @bobcatfish do we want to move ahead on this issue & PR?  Any outstanding questions or problems before we merge?

dibyom:
From my investigation in https://github.com/tektoncd/plumbing/issues/566, I think we should be able to move to a Deployment from a Stateful set with no issues.

boskos/boskos.yaml
```diff
@@ -91,31 +73,19 @@ spec:
       - name: boskos
         image: gcr.io/k8s-prow/boskos/boskos:v20200117-6384054e3
         args:
-        - --storage=/store/boskos.json
```
dibyom, line 94: This would have prevented https://github.com/tektoncd/plumbing/issues/566

sbwsg (approved):

sbwsg:
@tomgeorge feel free to cancel the hold if you want to proceed with this PR!

tomgeorge:
rebased master onto boskos-statefulset-to-deployment

Issue #206, opened 2020-01-30
Add a Tekton based check to enforce a `kind` label


afrittoli:
# Expected Behavior
To help release automation, it should not be possible to merge a PR if it doesn't have a "kind" label set. 

# Implementation plan
Create a new trigger that listens for PR changes.
Create a task that fetches the list of labels, if zero or more then one are found the check is failed.
Update the PR from the task with check passed or failed - and instructions on how to add a label if it's missing.

bobcatfish:
Failed comment could provide docs on what options can be used:

https://github.com/tektoncd/plumbing/blob/5ad2abad9a33c16496c9c2bce6a4d2a6db2d7250/label_sync/labels.yaml#L81-L122

bobcatfish:
I think this was fixed via https://github.com/tektoncd/plumbing/pull/229 anyway!

afrittoli:
> I think this was fixed via #229 anyway!

Indeed!

PR #229, opened 2020-02-11
Add a check that a PR has a kind label

tomgeorge:
# Changes

This PR adds a TriggerTemplate, TriggerBinding, and EventListener to check that the labels of a PR has a kind/ label.  If it has 0 or > 1 kind/ labels, it posts a comment saying that it cannot be merged in it's current state.
    
This enables #206 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/checks/check-pr-has-kind-label.yaml
+++ b/tekton/ci/checks/check-pr-has-kind-label.yaml
@@ -1,0 +1,219 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: check-labels-trigger
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["tekton.dev"]
+  resources: ["eventlisteners", "triggerbindings", "triggertemplates"]
+  verbs: ["get", "list"]
+- apiGroups: ["tekton.dev"]
+  resources: ["pipelineruns", "pipelineresources", "taskruns"]
+  verbs: ["create"]
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: gh-pr-label-checker
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: gh-pr-label-checker-rolebinding
+subjects:
+- apiGroup: ""
+  kind: ServiceAccount
+  name: gh-pr-label-checker
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: check-labels-trigger
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: trigger-to-check-pr-labels
+spec:
+  params:
+  - name: labels
+    value: $(body.pull_request.labels)
+  - name: gitRepository
+    value: $(body.repository.html_url)
+  - name: pullRequestUrl
+    value: $(body.pull_request.html_url)
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: github-pull-request-label-checker
+spec:
+  serviceAccountName: gh-pr-label-checker
+  triggers:
+    - name: trigger-check-pr-labels
+      interceptors:
+        - cel:
+            filter: >
+              header.match('X-GitHub-Event', 'pull_request') &&
+              (body.action == 'opened' ||
+              ((body.action == 'labeled' || body.action == 'unlabeled')
+              && body.label.name.contains('kind')))
+      bindings:
+        - name: trigger-to-check-pr-labels
+      template:
+        name: check-pull-request-labels
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerTemplate
+metadata:
+  name: check-pull-request-labels
+spec:
+  params:
+  - name: labels
+    description: List of labels currently on the Pull Request
+    type: string
+  - name: gitRepository
+    type: string
+  - name: pullRequestUrl
+    type: string
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1beta1
+    kind: TaskRun
+    metadata:
+      name: check-pr-labels-$(uid)
+    spec:
+      taskSpec:
+        resources:
+          inputs:
+            - name: source
+              type: git
+            - name: pr
+              type: pullRequest
+          outputs:
+            - name: pr
+              type: pullRequest
+        params:
+        - name: labels
+          description: The labels attached to the Pull Request
+        - name: gitRepository
+          description: The URL of the git repository
+        - name: pullRequestUrl
+          description: The URL of the Pull Request
+        volumes:
+        - name: label-config-v2
+          configMap:
+            name: label-config-v2
+        steps:
+        - name: copy-pr-to-output
+          image: busybox
+          script: |
+            #!/bin/sh
+            mkdir -p $(outputs.resources.pr.path)
+            cp -r $(inputs.resources.pr.path)/* $(outputs.resources.pr.path)/
+        - name: install-pyyaml
+          image: python:3-alpine
+          script: |
+            pip install pyyaml --user
+        - name: check-labels
+          image: python:3-alpine
+          volumeMounts:
+            - name: label-config-v2
+              mountPath: /etc/config
+          script: |
+            #!/usr/bin/env python
+
+            import json
+            import yaml
+
+            prLabels = json.loads('$(params.labels)')
+            labelNames = list(map(lambda e: e["name"], prLabels))
+            kindLabels = list(filter(lambda e: "kind" in e, labelNames))
+
+            availableLabels = None
+            with open("/etc/config/labels.yaml", "r") as stream:
+              availableLabels = yaml.safe_load(stream)["default"]["labels"]
+
+            availableKindLabels = list(filter(lambda e: "kind/" in e["name"], availableLabels))
+            availableKindNamesAndDescriptions = map(lambda e: "`" +str(e["name"])+ "`" + ":  " + str(e["description"]), availableKindLabels)
+
+            comment_template=""
+            if (len(kindLabels) > 1 or len(kindLabels) == 0):
+              comment_template += """
+            **This PR cannot be merged:** expecting exactly one kind/ label
+
+            <details>
+
+            Available `kind/` labels are:
+
+            """
+
+              for i in availableKindNamesAndDescriptions:
+                comment_template += i + "\n"
+
+              comment_template += """
+
+            </details>
+            """
+              new_comment_path = "$(outputs.resources.pr.path)/comments/new.json"
+              comment_body = dict(body=comment_template)
+              with open(new_comment_path, "w") as comment:
+                json.dump(comment_body, comment)
+
+              fail_check = {
+                "State": "failure",
+                "Label": "check-pr-has-kind-label",
+                "Desc": "The PR requires exactly one kind/ label to merge"
+              }
+
+              new_check_path = "$(outputs.resources.pr.path)/status/check-pr-has-kind-label.json"
+              with open(new_check_path, "w") as check:
+                json.dump(fail_check, check)
+            else:
+              success_check = {
+                "State": "success",
+                "Label": "check-pr-has-kind-label",
+                "Desc": "This PR has exactly one kind/ label"
+              }
+              new_check_path = "$(outputs.resources.pr.path)/status/check-pr-has-kind-label.json"
+              with open(new_check_path, "w") as check:
+                json.dump(success_check, check)
+      inputs:
+        resources:
+        - name: source
+          resourceSpec:
+            type: git
+            params:
+            - name: revision
+              value: master
+            - name: url
+              value: $(params.gitRepository)
+        - name: pr
+          resourceSpec:
+            type: pullRequest
+            params:
+            - name: url
+              value: $(params.pullRequestUrl)
+            secrets:
+            - fieldName: authToken
+              secretName: github-token
+              secretKey: oauth
+        params:
+        - name: labels
+          value: $(params.labels)
+        - name: gitRepository
+          value: $(params.gitRepository)
+        - name: pullRequestUrl
+          value: $(params.pullRequestUrl)
+      outputs:
+        resources:
+        - name: pr
+          resourceSpec:
+            type: pullRequest
+            params:
+            - name: url
+              value: $(params.pullRequestUrl)
+            secrets:
+            - fieldName: authToken
+              secretName: github-token
+              secretKey: oauth


tomgeorge:
I can't seem to make the check show up properly.  On the first invocation, I can see the check fail.  But on subsequent runs, if the check succeeds and then fails, it stays green.

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: triggers-minimal
```
afrittoli: We might factor up roles and service accounts for triggers at some point, but until then we'll have to keep the name unique enough to avoid conflicts

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+    type: string
+  - name: gitRepository
+    type: string
+  - name: pullRequestUrl
+    type: string
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1alpha1
+    kind: PipelineResource
+    metadata:
+      name: pr-$(uid)
```
afrittoli: My initial though was that we could use a name that helps identifying the source, like label-check-pr-$(uid). 
Trigger adds labels though, so we can use those to filter resources associated to a check, e.g.

```
  labels:
    tekton.dev/eventlistener: dashboard-nightly-release-cron
    tekton.dev/trigger: dashboard-nightly-release-cron-trigger
```

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+    metadata:
+      name: git-source-$(uid)
+    spec:
+      type: git
+      params:
+      - name: revision
+        value: master
+      - name: url
+        value: $(params.gitRepository)
+      - name: labels
```
afrittoli: Is "labels" a valid parameter for a git resource? Why do we need the PR labels on the git resource?

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+        steps:
+        - name: copy-pr-to-output
+          image: busybox
+          script: |
+            #!/bin/sh
+            mkdir -p $(outputs.resources.pr.path)
+            cp -r $(inputs.resources.pr.path)/* $(outputs.resources.pr.path)/
+        - name: check-labels
+          env:
+          - name: GH_LABELS
```
afrittoli: Unused?

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+            comment_template=""
+            if (len(kindLabels) > 1 or len(kindLabels) == 0):
+              print('failed')
+              comment_template += """
+            **This PR cannot be merged:** expecting one kind/ label
+
+            <details>
+            The following `kind/` labels are available:</p>
+
+            These are some kind labels
```
afrittoli: This list should ideally be pulled from the configuration in the plumbing repo, so that it's always up to date.

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+              success_check = {
+                "State": "success",
+                "Label": "check-pr-has-kind-label",
+                "Desc": "This PR has a kind/ label"
+              }
+              new_check_path = "$(outputs.resources.pr.path)/status/check-pr-has-kind-label.json"
+              with open(new_check_path, "w") as check:
+                json.dump(success_check, check)
+        - name: cat-check
+          image: ubuntu
```
afrittoli: It may be enough to use a smaller image just to run a `cat` :)

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,203 @@
...
+          script: |
+             #!/bin/bash
+             cat $(outputs.resources.pr.path)/status/*
```
afrittoli: This is for debug purposes only? 
Do we need to log *all* checks?

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: This was a debugging step that I left in.  I will remove it after I get everything set up

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Correct, this is for debugging

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: I pass in the labels from the event so that I can check their contents.  Is there a better spot that I should grab them from?

tomgeorge: Oh I see, it doesn't belong there.  Removed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: This has been fixed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Fixed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Removed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Removed

tomgeorge:
I think this is blocked by https://github.com/tektoncd/pipeline/issues/2062.  We could forgo the check for now, and just add a comment and a do-not-merge label.

tomgeorge:
@afrittoli thanks for the status PR!  This should work now

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,218 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: check-labels-trigger
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["tekton.dev"]
+  resources: ["eventlisteners", "triggerbindings", "triggertemplates", "tasks", "taskruns"]
```
dibyom: The tasks and taskruns resources here are not necessary

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,218 @@
...
+  - name: labels
+    description: List of labels currently on the Pull Request
+    type: string
+  - name: gitRepository
+    type: string
+  - name: pullRequestUrl
+    type: string
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1alpha1
+    kind: PipelineResource
```
dibyom: It might be worth in lining/embedding these resources as resourceSpecs in the TaskRun below: https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md#best-practices

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Thanks, changed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Done!

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+    value: $(body.pull_request.html_url)
+---
+apiVersion: tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: github-pull-request-label-checker
+spec:
+  serviceAccountName: gh-pr-label-checker
+  triggers:
+    - name: trigger
```
afrittoli: The trigger name is attached to the resources provisioned by the template in a label, which is very useful for filtering taskruns. Would you mind using a trigger name that is more indicative here?
I know we used "trigger" in other places and we need to fix that.

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+      interceptors:
+        - cel:
+            filter: "header.match('X-GitHub-Event', 'pull_request') && (body.action == 'opened' || ((body.action == 'labeled' || body.action == 'unlabeled') && body.label.name.contains('kind')))"
```
afrittoli: NIT: This might be more readable as a multi-line - but the filter looks ok.

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+        - name: copy-pr-to-output
+          image: busybox
+          script: |
+            #!/bin/sh
+            mkdir -p $(outputs.resources.pr.path)
+            cp -r $(inputs.resources.pr.path)/* $(outputs.resources.pr.path)/
+        - name: install-pyyaml
+          image: python:3-alpine
+          script: |
+            pip install pyyaml --user
```
afrittoli: Does this work? 
I guess `pip install pyyaml --user` downloads it into `/home/tekton` which stays for the next container.
I would prefer not to install packages as part of the CI job.

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+            prLabels = json.loads('$(params.labels)')
+            labelNames = list(map(lambda e: e["name"], prLabels))
+            kindLabels = list(filter(lambda e: "kind" in e, labelNames))
+
+            availableLabels = None
+            with open("/etc/config/labels.yaml", "r") as stream:
+              availableLabels = yaml.safe_load(stream)["default"]["labels"]
+
+            availableKindLabels = list(filter(lambda e: "kind/" in e["name"], availableLabels))
+            availableKindNamesAndDescriptions = map(lambda e: "`" +str(e["name"])+ "`" + ":  " + str(e["description"]), availableKindLabels)
```
afrittoli: Since we already enforce the list of available labels, I don't believe we need to do this. It should be enough to check that  a label that starts with "kind/" in the list of labels attached to the PR.
@vdemeester thoughts?

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+            """
+              new_comment_path = "$(outputs.resources.pr.path)/comments/new.json"
+              comment_body = dict(body=comment_template)
+              with open(new_comment_path, "w") as comment:
+                json.dump(comment_body, comment)
+
+              fail_check = {
+                "State": "failure",
+                "Label": "check-pr-has-kind-label",
+                "Desc": "The PR requires exactly one kind/ label to merge"
```
afrittoli: NIT: We might want to add a text with a sample command to add the label, e.g. "/kind [feature|documentation|...]"
but that's something we can also add on top

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,216 @@
...
+  - name: pullRequestUrl
+    value: $(body.pull_request.html_url)
+---
+apiVersion: tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: github-pull-request-label-checker
+spec:
+  serviceAccountName: gh-pr-label-checker
+  triggers:
```
afrittoli: We will need to have `serviceType: NodePort` to expose this via an ingress eventually, but this is something we can do in a later patch.

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Changed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Changed

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: What if the label starts with `kind/` but isn't in `labels.yaml`?

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Yes this does work!  I am using it parse `labels.yaml` in the next step.  If we don't need it then I will remove it, but it seems like we do.

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: When the check fails, the issue comment includes a list of possible `kind/` labels to add.

tekton/resources/checks/check-pr-has-kind-label.yaml
afrittoli: That should not be possible

tekton/resources/checks/check-pr-has-kind-label.yaml
afrittoli: oh, nice

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,220 @@
...
+              header.match('X-GitHub-Event', 'pull_request') && 
+              (body.action == 'opened' || 
+              ((body.action == 'labeled' || body.action == 'unlabeled') 
+              && body.label.name.contains('kind')))
+      bindings:
+        - name: trigger-to-check-pr-labels
+      template:
+        name: check-pull-request-labels
+---
+apiVersion: tekton.dev/v1alpha1
```
afrittoli, line 67: the api group changes to `triggers.tekton.dev` in triggers v0.4.0 - which we now run - could you update all trigger resources accordingly?

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,220 @@
...
+  params:
+  - name: labels
+    description: List of labels currently on the Pull Request
+    type: string
+  - name: gitRepository
+    type: string
+  - name: pullRequestUrl
+    type: string
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1alpha1
```
afrittoli, line 81: this is now v1beta1

tekton/resources/checks/check-pr-has-kind-label.yaml
```diff
@@ -0,0 +1,220 @@
...
+    kind: TaskRun
+    metadata:
+      name: check-pr-labels-$(uid)
+    spec:
+      taskSpec:
+        inputs:
```
afrittoli, line 87: when changing to v1beta1, `resources` and `params` become root keys in the spec

tekton/resources/checks/check-pr-has-kind-label.yaml
tomgeorge: Is this ok to keep then, if `availableKindNamesAndDescriptions` is printed out in the GH comment when the check does not pass?

tomgeorge:
made the api version changes, moved into `tekton/ci/checks` and squashed into one commit

tomgeorge:
ok let me try

Issue #159, opened 2019-12-13
Migrate tektoncd project to Tekton (instead of prow)


vdemeester:
This is kind of an Epic issue for moving towards having Tekton building Tekton.
The ultimate goal is to have all CI/CD (aka PR builds, periodics builds, release builds, commands handling, …) of the `tektoncd` projects handled by `tektoncd` projects (aka `tektoncd/pipeline`, `tektoncd/triggers`, …).

bobcatfish:
These are the issues in the epic (retrieved from zenhub):

- plumbing#102 - Tektoncd project convention for pipeline-as-code
- plumbing#103 - Towards tektoncd project with no dependencies to plumbing
- plumbing#124 - Make the Mario bot production ready
- plumbing#137 - Automate obtaining a test cluster
- plumbing#158 - Setup periodics job for tektoncd projects (using dogfooding)
- plumbing#58 - Automated release for tektoncd projects

afrittoli:
As a first step, we shall maintain job definitions in the `tektoncd/plumbing` repo, similar to what we do today with prow.
Ultimately it would be nice move the definition to the respective repos, i.e. pipeline tests in `tektoncd/pipeline`.

vdemeester:
/area roadmap
This is probably a duplicate but…

PR #254, opened 2020-03-02
Add a concept for CI with Tekton

afrittoli:
# Changes

Initial set of resources. The CI pipeline is a no-op.
Resources are not separated for now, but the github
template, most bindings and conditions are already reusable.

The CI pipeline adds two CI Jobs. Each first asynchronously
resets the check status on the PR, adding a link to the
running pipelinerun. Then it runs the CI job and finally
it asynchronously updates the check status with the CI job
status and a link to the dashboard.
The first job always succeeds, the second fails 50% of the times,
which is handy to demo the /test facility. 

The execution of the two jobs is conditional. Both jobs
are attached to a condition that will check the content
of the PR and match it to a regular expression. If there
is a match the jobs is executed.

The regex for the failed job is: "(tekton/ci|tekton-demo)"

Add a trigger that handles the "/test" comment on pull request issues.
The event goes through github and cel interceptors for validation and
filtering. Since the comment event does not contain enough details
about the PR itself, the event also goes though a new custom
interceptor "add-pr-body" which fetches the PR details from GitHub
API and appends them to event body.

Related to #159

# Submitter Checklist

review them:


afrittoli:
I think this PR is close to be ready to go, but there's plenty of work needed before this is more than experimental:
- finalize the interface for CI Jobs and document what are the requirements for a Task/Pipeline so that it can be used as a CI Job
- setup the infra required by Jobs that need a k8s cluster to run
- document how to add new CI Jobs
- further split resources so that projects has close to zero boiler-plate YAML write when configuring their CI jobs

Nice to have soon:
- setup the infra required to collect and store test results
- setup a DB with runs (and test results) and dashboard on top
- use task results to transport dynamic metadata
- drop the need for having a cloudevent output resource in CI Tasks

tekton/ci/jobs.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: tekton-noop
+  namespace: tektonci
+  description: |
+    Example CI Job that does nothing interesting.
```
vdemeester: This is gonna become a field "soon-ish" :stuck_out_tongue:

afrittoli:
> If the build fails, does the github status stay pending ? 🤔
No, it will send a cloud event which contains the failed status, and the github status will be updated accordingly. I will add a second job that always fails to complete the demo and verify that this works as expected.

> Otherwise, we could try to use https://github.com/tektoncd/experimental/tree/master/commit-status-tracker for the github status 😛 👼
Yeah, possibly. I did not manage to look into that yet.

tekton/ci/bindings.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: tekton-ci-webhook-pull-request
+  namespace: tektonci
+spec:
+  params:
+  - name: buildUUID
+    value: $(header.X-GitHub-Delivery)
```
afrittoli: The problem with this is that all CI Jobs triggered by this event get the same buildUUID, so the logs get mixed up, like in http://logs.dogfooding.tekton.dev/?buildid=f46f3900-5edb-11ea-9990-2f5693adf309&namespace=tektonci
We could append something job-specific to the UUID, but that would break the format expected by the log app. Perhaps we could extend the log app to support that,

vdemeester: cc @sbwsg

sbwsg: Yeah, totally agree.  I'd be happy to update the log viewer app to accept a `taskName` or something to disambiguate the log output.

tekton/ci/README.md
```diff
@@ -0,0 +1,159 @@
...
+CloudEvent PipelineResource that points to the `el-github-check-done.tektonci`
+service.
+
+This is true until we will be able to trigger cloud events without the need of
+a CloudEvent PipelineResource.
+
+Example pipeline:
+
+```
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/conditions.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Condition
+metadata:
+  name: check-git-files-changed
+  namespace: tektonci
+  annotations:
+    description: |
+      Succeeds if any of the files changed in the last N commits from the
+      revision defined in the git resource matches the regular expression
```
vdemeester: On 0.11-rc2 this can go into the `description` field

tekton/ci/github-template.yaml
```diff
@@ -0,0 +1,101 @@
...
+    spec:
+      type: pullRequest
+      params:
+        - name: url
+          value: $(params.gitURL)/pull/$(params.pullRequestNumber)
+      secrets:
+        - fieldName: authToken
+          secretName: bot-token-github
+          secretKey: bot-token
+  - apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/jobs.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/jobs.yaml
```diff
@@ -0,0 +1,69 @@
...
+      #!/bin/sh
+      set -ex
+      echo "$(inputs.params.message)"
+      echo
+      sleep 120
+      echo "Content of the repo:"
+      ls $(inputs.resources.source.path)
+      [[ "$(inputs.params.passorfail)" == "pass" ]]
+---
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/plumbing-template.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - name: gitCloneDepth
+    description: Number of commits in the change + 1
+  - name: gitHubCommand
+    description: |
+      The GitHub command that was used a trigger. This is only available when
+      this template is triggered by a comment. The default value is for the
+      case of a pull_request event.
+    default: ""
+  resourcetemplates:
+  - apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/plumbing-template.yaml
```diff
@@ -0,0 +1,114 @@
...
+            value: $(params.gitRepository)
+          - name: depth
+            value: $(params.gitCloneDepth)
+      - name: starttrigger
+        resourceRef:
+          name: github-check-start
+      - name: endtrigger
+        resourceRef:
+          name: github-check-done
+  - apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/tasks.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/tekton-noop-check.yaml
```diff
@@ -0,0 +1,135 @@
+apiVersion: tekton.dev/v1alpha1
```
vdemeester: Should we need to migrate to `v1beta1` now ?

tekton/ci/README.md
afrittoli: Well, I have not deployed 0.11.0-rc1 to dogfooding yet.

vdemeester: ah ok :angel: Then later on :stuck_out_tongue:

shiyu-dong:
@bobcatfish Thanks for the update. This looks very promising! I will keep an eye on that plumbing folder :)

One question though. Based on this [diagram](https://github.com/tektoncd/plumbing/blob/master/tekton/ci/docs/ci-concept.md) seems that all the triggers are more like "fire and forget" and N `GitHub Updater Tasks` will be triggered eventually, one for each child pipeline.

What if I want to mark the parent CI job as successful only when all the N child pipeline runs are successful? In other words can we "register" the N child pipeline runs to the second trigger before they start, so that the `GitHub Updater Task` will be triggered only when the cloud events for all these N child pipelines are received?

Just my two cents, I do have some ideas around extending the PipelineRun CRD to achieve the same thing. Today a TaskRun is "static" (1 run instance per task definition), but we can make it "dynamic" (N run instances per task definition), and only consider this TaskRun as finished when all the N instances finishes (either success or fail). We can somehow define a `flatten` operation on the parameter (**list** of size N) of the dynamic task, and for each run instance, use one entry from the list as the parameter to start the run. In this way other PipelineRun orchestration (DAG, failure strategy, etc) can work as is, everything is k8 native and there's no RPC overhead with using triggers.

Frankly this is a hard problem to solve and very few CI system can do this. It will be great if that can be something natively supported by Tekton in an easy way.

Issue #224, opened 2020-02-07
Schedule and document CD of pipeline, triggers and dashboard on "robocat"


afrittoli:
# Expected Behavior
Tekton services pipeline, triggers and dashboard are deployed nightly from the respective nightly build. The process is documented in the "robotcat" setup document. 

The pipeline for deployment from a release with overlays already exists. 
Dedicated overlays need to be created for the "robocat" cluster.

Any prerequisite to the deployment must also be automated, documented and deployed regularly.

Steps:
1. Add a cadmin account for deployment root resources and a tekton deployer account for deployment of services. Refactor a bit the CD trigger, collapse shared bindings and separate them into parts based on functionality. Start running deployment of "cadmin" resource to robcat on a nightly basis.
2. Implement the ability to retrieve the latest nightly release of a service. Schedule nightly deploys of pipeline, triggers and dashboard, with the ability to pin a specific version and source bucket

afrittoli:
This is fixed now. Nightly releases are deployed to robocat, and the deployment is documented.

PR #270, opened 2020-03-12
Implement CD services for robocat cluster

afrittoli:
# Changes

Refactor the CD event listener so that we only use one that
supports different types of CD: "configmaps" and "folders" for now.
Aggregate and then functionally split bindings for CD.
Adapt existing cronjobs (prow and label sync) to the new resource
layout.

Add a cadmin account for deployment root resources and a tekton
deployer account for deployment of services in a "cadmin" folder.

Start running deployment of "cadmin" resource to robcat on a nightly basis.

Work on #224

# Submitter Checklist

review them:

for more details._


robocat/cadmin/200-rbac.yaml
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cadmin
+  namespace: default
```
bobcatfish: i see we have `cadmin` in `tektoncd` also - this one in `default` doesnt seem to get any bindings, do we need it?

bobcatfish: im also not completely clear on the difference in the intended usage for` tekton-deployer` vs `cadmin` (i dont see `tekton-deployer` in the docs - maybe it's in some docs that already exist?

tekton/resources/cd/clusters.yaml
```diff
@@ -37,3 +50,43 @@ spec:
...
+  - name: username
+    value: cadmin
+  secrets:
+  - fieldName: token
+    secretKey: token
+    secretName: robocat-tektoncd-cadmin-token
+  - fieldName: cadata
+    secretKey: ca.crt
+    secretName: robocat-tektoncd-cadmin-token
+  type: cluster
```
bobcatfish: oooo @sbwsg we should make that cluster Task asap :D

tekton/cronjobs/robocat-cadmin-cron/cronjob.yaml
```diff
@@ -0,0 +1,26 @@
...
+    spec:
+      template:
+        spec:
+          containers:
+          - name: trigger
+            env:
+              - name: GIT_REPOSITORY
+                value: "github.com/afrittoli/plumbing"
+              - name: GIT_REVISION
+                value: "issues/224"
```
bobcatfish: are these values temporary? i guess you want to update them after the merge?

(we need to get some kinda way setup to apply changes that are within a pull request more smoothly one day! :D)

tekton/cronjobs/prow-config-cd-hourly-cron/cronjob.yaml
```diff
@@ -24,3 +22,9 @@ spec:
                 value: "default"
               - name: CLUSTER_RESOURCE
                 value: "prow-cluster-config-bot"
+              - name: CONFIGMAP_NAME
+                value: "config"
+              - name: CONFIGMAP_KEY
+                value: "config.yaml"
+              - name: CONFIGMAP_DESCRIPTION
+                value: "prow-config"
```
bobcatfish: neat!!! :D nice improvement

tekton/resources/cd/eventlistener.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - name: deploy-target-details
+        - name: deploy-source-git
+        - name: configmap-details
+      template:
+        name: deploy-configmap
+    - name: folders
+      interceptors:
+        - cel:
+            filter: >-
+              'folder' in body
```
bobcatfish: could you explain more about what these filters are looking for (maybe in a comment?)

im a bit confused what the contents of the body would be here if we're triggering with a cron and why sometimes they'd contain configmap, sometimes folder, and why thats enough info to filter on

tekton/resources/cd/folder-template.yaml
```diff
@@ -0,0 +1,106 @@
...
+spec:
+  params:
+  - name: gitRepository
+    description: URL of the repository that holds the folder
+  - name: gitRevision
+    description: Git revision
+  - name: namespace
+    description: Namespace to deploy to in the target cluster
+  - name: clusterResource
+    description: Name of the cluster resource that points to the target cluster
```
bobcatfish: interesting, i never imagined anyone using a PipelineResource like this - not sure our currently plan with Tasks takes this kinda usage into account ( + @sbwsg )

tekton/resources/cd/folder-template.yaml
```diff
@@ -0,0 +1,106 @@
...
+            [[ "$(inputs.params.isOverlay)" == "true" ]] && METHOD_PARAM="-k"
+
+            # Determine whether to enforce namespace across resources
+            NAMESPACE_PARAM="-n $(inputs.params.namespace)"
+            [[ "$(inputs.params.namespace)" == "" ]] && NAMESPACE_PARAM=""
+
+            # Run the deployment
+            kubectl "$(inputs.params.deployMethod)" \
+              $NAMESPACE_PARAM $METHOD_PARAM \
+              $(inputs.resources.source.path)/$(inputs.params.folderPath)
```
bobcatfish: this seems like a totally reasonable amount of scripting to me :D v nice!

robocat/cadmin/200-rbac.yaml
afrittoli: oh, nice catch, thank you - I moved this to root and forgot to remove it from here
'tekton-deployer' is the admin of 'tekton-pipelines' so it will be used to deploy tekton services. I could have add this in the next PR, sorry

tekton/cronjobs/robocat-cadmin-cron/cronjob.yaml
afrittoli: upz, yeah, they were not meant to be added in the PR - I was just trying to test the setup in the dogfooding repo.

afrittoli: > (we need to get some kinda way setup to apply changes that are within a pull request more smoothly one day! :D):

+100 - that's the direction we want to go for sure!

tekton/resources/cd/eventlistener.yaml
afrittoli: This is a "fan out" that is used by the event listener to trigger the right type of deployment.
It basically allows having one entry point for deployment jobs, which can be used to deploy config maps, or resources from a folder or other types in future, based on the parameters sent in the body.

The alternative would be to have multiple event listeners - which may be a valid one.
It seemed to me that having one running service only reduces a bit boiler plate code and makes it more manageable, but I don't have a very strong feeling about this.

tekton/resources/cd/folder-template.yaml
afrittoli: I still like very much the abstraction layer that is given by pipeline resources. 
But there are alternatives for this kind of setup as well.
You could pass around the parameters of the pipeline resource and generate the kubeconfig adding the catalog task at the beginning of the pipeline.
However I feel that the pipeline resource allows to keep the params grouped together and to specify a clearer interface.

tekton/resources/cd/folder-template.yaml
afrittoli: Heh, this is a poor man version of a proper deployment system. There are several things that are not handled in here, like resource renaming and deletion for instance to name a few. I started looking at `flux` to see if that is something that we could integrate in here.

afrittoli:
The commit in this PR was now merged as part of #271 .
I tried to make a stack of PR to split the work on #224, but there's no good way in GH to express dependencies between GitHub. Perhaps we shall build a Tekton based service that handles "depends-on" type of comments :)

vdemeester:
@afrittoli oh lol.. I thought the dependency was the way around.. I `lgtm`ed the wrong branch then :D

afrittoli:
https://github.com/z0al/dep 😮

PR #281, opened 2020-03-23
Update build bot to load rotation dynamically 🚚

bobcatfish:
# Changes

Build cop bot will now load the csv file from github every time it checks it.

Previously the buildbot binary was being built with the rotation csv as
part of the binary itself, which meant that updating the rotation would
require rebuilding and redeploying the image.

This change makes it so that the rotation can be updating without
needing to update the image - the downside is that the running binary is
now coupled to the repo but I think it's worth it!

Follow up TODOS:
* run unit tests on every PR https://github.com/tektoncd/plumbing/issues/283
* deploy build bot binary automatically https://github.com/tektoncd/plumbing/issues/282

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/buildbot/cmd/buildbot/main.go
+++ b/buildbot/cmd/buildbot/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/csv"
 	"fmt"
 	"log"
 	"os"
@@ -19,21 +18,7 @@ var (
 	channelID  string
 
 	vdemeest string
-	cops     = map[string]struct{}{}
-	rotation = map[string]string{}
 )
-
-func init() {
-	// When building with "ko", templates is deployed under KO_DATA_PATH
-	// If KO_DATA_PATH is not defined, the path will be a relatove one
-	basePath := os.Getenv("KO_DATA_PATH")
-	csvfile := path.Join(basePath, "rotation.csv")
-	if r, err := readRotation(csvfile); err != nil {
-		panic("Cannot read rotation.csv… bailing 😓")
-	} else {
-		rotation = r
-	}
-}
 
 func main() {
 	token = os.Getenv("SLACKTOKEN")
@@ -62,16 +47,19 @@ func main() {
 		if user.Name == "vdemeest" {
 			vdemeest = user.ID
 		}
-		if _, ok := cops[user.Name]; ok {
-			copsID[user.Name] = user.ID
-		}
+		copsID[user.Name] = user.ID
 	}
 
-	currentCop = getBuildCop(time.Now())
+	// When building with "ko", templates is deployed under KO_DATA_PATH
+	// If KO_DATA_PATH is not defined, the path will be a relative one
+	basePath := os.Getenv("KO_DATA_PATH")
+	csvfile := path.Join(basePath, "rotation.csv")
+	r := NewRotation(csvfile)
+	currentCop = copsID[r.GetBuildCop(time.Now())]
 
 	rtm := api.NewRTM()
 	go rtm.ManageConnection()
-	go dailyPing(rtm, copsID)
+	go dailyPing(rtm, copsID, r)
 
 	for msg := range rtm.IncomingEvents {
 		switch ev := msg.Data.(type) {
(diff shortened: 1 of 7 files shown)


sbwsg:
lol i forgot cant approve

/lgtm

buildbot/cmd/buildbot/main.go
```diff
@@ -62,16 +48,15 @@ func main() {
...
 			vdemeest = user.ID
 		}
-		if _, ok := cops[user.Name]; ok {
-			copsID[user.Name] = user.ID
-		}
+		copsID[user.Name] = user.ID
 	}
 
-	currentCop = getBuildCop(time.Now())
+	r := NewRotationFromURL(rotationURL)
```
sbwsg: If we copied this into `dailyPing`'s for loop, would we also get updates pulled in every day to the csv?  Any reason not to do that?

sbwsg: nm!  I realised that calling r.GetBuildCop(time.Now()) makes a request for a new version of the CSV.  Nice!

buildbot/cmd/buildbot/rotation.go
```diff
@@ -0,0 +1,83 @@
...
+// the file getter f it is initialized with.
+type Rotation struct {
+	f GetFile
+}
+
+// GetFile is the signature of a function that knows how to retrieve the bytes from a file
+type GetFile func() (io.ReadCloser, error)
+
+// NewRotationFromFile returns a new Rotation object which will read from file.
+func NewRotationFromFile(file string) Rotation {
```
vdemeester: Why not the following :

```go
func NewRotation(readFn GetFile) Rotation {
    // […]
}

func FromFile(file string) GetFile {
    return func() (io.ReadCloser, error) { … }
}

func FromURL(url string) GetFile …
```

vdemeester: Also, this is only used for the tests right, so… it should only be defined there :stuck_out_tongue_winking_eye:

buildbot/cmd/buildbot/rotation.go
```diff
@@ -0,0 +1,83 @@
...
+		return resp.Body, nil
+	}}
+}
+
+// GetBuildCop returns the name of the build cop for the requested time
+func (r Rotation) GetBuildCop(t time.Time) string {
+	tf := t.Format("2006-01-02") // Mon Jan 2 15:04:05 MST 2006
+	f, err := r.f()
+	if err != nil {
+		log.Errorf("Could not read from build cop rotation: %v", err)
```
vdemeester: Any reason to use `log` here instead of `fmt` ?
Also, we should *most likely* use the built-in log instead of `k8s.io/kubernetes/pkg/kubelet/kubeletconfig/util/log"`, as.. bringing k8s.io/kubernetes dependency just for logs is a high coste :angel:

buildbot/cmd/buildbot/rotation.go
bobcatfish: I guess so - it only _happens_ to be used by the unit tests, but it's a pretty reasonable part of this overall package

vdemeester: :stuck_out_tongue_closed_eyes: true, but it ain't use by anything except the test, it doesn't need to be exported :angel: 
If anybody needs this to be exported at some point, a PR can be opened :angel: 

(that's usually my take on things : whatever doesn't need to be exported *right now*, shouldn't be exported)

bobcatfish:
> lol i forgot cant approve

https://github.com/tektoncd/plumbing/blob/master/OWNERS *whispers* _jooooooin us @sbwsg_

buildbot/cmd/buildbot/rotation.go
bobcatfish: oooo i like it!

bobcatfish:
> This is not super efficient and ties us to the GitHub status 😅 which is kinda fine as if GitHub is down, there is less buildcop to do, but…

Definitely a bit less efficient! Whaddaya think @vdemeester do you want to not do it? Alternatively:
* do this for now
* go back to using bundled csv file after https://github.com/tektoncd/plumbing/issues/282?

truthfully i think #282 is the better solution but i was intimidated by the lack of other automation on this repo and unsure where to start so i went with a shorter term fix in the meantime 🙏

buildbot/cmd/buildbot/rotation.go
bobcatfish: kk ill try it! i spose it makes sense to expose as little as possible, that sounds like a principle i learned a while ago or something 🤔

buildbot/cmd/buildbot/rotation.go
bobcatfish: oh dear i did NOT mean to bring that in! that explains the go.mod change 🤦‍♀

bobcatfish: i meant to use the built in `log` lib, which i thought made a bit more sense than `fmt` but im open to either!

vdemeester: ah lol.. I'll wait for your follow-up :stuck_out_tongue_closed_eyes:

PR #287, opened 2020-03-31
Prerelease Check: Only fail if version tag endings are exact match

sbwsg:
# Changes

Previously the prerelease check would fail if any tag in the git repo _contained_ the version tag supplied to the Task.  For example, `v0.11.0` would be disallowed if there was an existing tag of `v0.11.0-rc1`.

This PR changes the prerelease check so that the tag endings must match exactly.  `v0.11.0` will be allowed even if tag `v0.11.0-rc1` exists.

# Submitter Checklist

review them:

for more details._


sbwsg:
One data point so far: I used this in the release of 0.11.0 and it worked for me.  Before it was stopping at prerelease check because of the existing 0.11.0-rcX tags.

PR #296, opened 2020-04-03
Improve tekton-experimental-helm-lint prow job

eddycharly:
Use kubernetes agent instead of pipeline.

This is related to tektoncd/experimental#494


eddycharly:
Working towards #306.

PR #302, opened 2020-04-08
Add decorate flag on pull-tekton-experimental-helm-lint

eddycharly:
Added `decorate: true` on `pull-tekton-experimental-helm-lint` prow job.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -456,6 +456,7 @@ presubmits:
     always_run: true
     rerun_command: /run tekton-experimental-helm-lint
     trigger: "(?m)^/run (all|tekton-experimental-helm-lint),?(\\s+|$)"
+    decorate: true
     spec:
       containers:
       - image: alpine/helm:3.1.2


bobcatfish:
It's too late to do this in the commit message, but @eddycharly (or @vdemeester if you know) could you explain a bit more about why this flag is needed? i.e. what this adds, and was something broken without it?

Thanks!

eddycharly:
@bobcatfish it adds decoration to the pod running the job.
That is, it adds some other containers that are responsible of cloning the repository, setting the correct working dir, uploading logs and artifacts etc...

bobcatfish:
Thanks @eddycharly that helps! I just noticed https://github.com/tektoncd/plumbing/pull/305 as well - i mentioned there but having an issue or something that these PRs could link to would help provide some context

I know it's hard to get these prow things working first shot since it's hard to try them out :(

eddycharly:
You can read more here: https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md

I wrote the job for pipeline agent first but i reverted to kubernetes agent while thinking the pipeline agent was the issue.
In the end the issue was not directly related to the agent, i could try to use pipeline agent again but i fear logs would not be uploaded.

bobcatfish:
Thanks @eddycharly that helps too!

What I mean though is the context for adding pull-tekton-experimental-helm-lint which I assume you are working toward adding with these PRs? (Basically I have to guess the context for this work at the moment, maybe I'm wrong)

eddycharly:
Working towards #306.

@bobcatfish i created the issue, linking it in the PRs for history.

PR #319, opened 2020-04-10
Add Task and Triggers for deploying buildbot

dibyom:
# Changes

Current State:
* Task that runs `ko apply` 

TODO:
- [ ] Use workspaces instead of volumes for GCP secret or use Workload Identity
- [ ] Add a CronJob Trigger that runs the above task every day
- [ ] Document the Secret/ServiceAccount roles/permissions

Maybe TODO:
- [ ] Use Workspace/git clone task instead of PipelineResource

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/buildbot/tekton/deploy.yaml
+++ b/buildbot/tekton/deploy.yaml
@@ -1,0 +1,37 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: buildbot-deploy
+spec:
+  resources:
+    inputs:
+    - name: source
+      type: git
+      targetPath: go/src/github.com/tektoncd/plumbing
+  params:
+  - name: imageRegistry
+    default: "gcr.io/dibyo-spinnaker-dev/pipeline"
+  steps:
+  - name: buildbot-build-push-deploy
+    image: gcr.io/tekton-nightly/ko-ci
+    env:
+    - name: KO_DOCKER_REPO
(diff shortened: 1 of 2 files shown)


buildbot/tekton/deploy.yaml
```diff
@@ -0,0 +1,37 @@
...
+      value: /workspace/go
+    - name: CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE
+      value: /secret/release.json
+    script: |
+      #!/bin/sh
+      set -ex
+      gcloud auth configure-docker
+
+      cd /workspace/go/src/github.com/tektoncd/plumbing/buildbot
+      ko apply -f config/
```
afrittoli, line 30: Thanks for this!

This will not improve the current situation where it's hard to track what version is deployed.
My suggestion for this would be to:
- either add a nightly build of all the bots (like we do for pipeline/triggers/dashboard) and install the nightly
- or at least use the -t (for tag) option of ko, so that we tag the resulting image with something traceable, i.e. either a git sha or a daily tag like we use for nightly builds

I think the latter would be easy to implement, and we can still extend to the first option later if we feel it's needed

dibyom: Both good ideas. I was planning on adding a CronJob that would deploy it on a regular basis. Will add the tags as well!

Issue #326, opened 2020-04-15
Postsubmit prow job logs are not uploaded at the correct location


eddycharly:
# Expected Behavior

Prow postsubmit job logs are available from github.

# Actual Behavior

Prow postsubmit job logs are uploaded at a location that is not consistent with the prow configuration.

# Steps to Reproduce the Problem

Take any commit on a repo, show commit checks.

![image](https://user-images.githubusercontent.com/47974576/79364259-6742e980-7f49-11ea-8bec-a26fc3a1add6.png)

Click on details, the logs are not found.

![image](https://user-images.githubusercontent.com/47974576/79364564-d0c2f800-7f49-11ea-9285-5811599f7bbc.png)

# Additional Info

`pr-logs` should be replaced by `logs` at the locations below:

https://github.com/tektoncd/plumbing/blob/ba56b3c3a78bad79177b4730826d780ddcf36319/prow/config.yaml#L1082

https://github.com/tektoncd/plumbing/blob/ba56b3c3a78bad79177b4730826d780ddcf36319/prow/config.yaml#L1113

https://github.com/tektoncd/plumbing/blob/ba56b3c3a78bad79177b4730826d780ddcf36319/prow/config.yaml#L1159

eddycharly:
Fixed, see here for example: https://tekton-releases.appspot.com/build/tekton-prow/logs/tekton-pipeline-publish-images/1250534403294105601/

PR #327, opened 2020-04-15
Fix postsubmit logs location

eddycharly:
Fixing #326

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1079,7 +1079,7 @@ postsubmits:
         - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
         - "--root=/go/src"
         - "--service-account=/etc/nightly-account/service-account.json"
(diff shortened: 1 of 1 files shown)


sbwsg:
This looks sound to me but I have to confess I don't actually know if it's correct or not.  @dibyom, @afrittoli, or @vdemeester might know more?

eddycharly:
It is correct, in fact it can be anything but it has to be consistent with the config here:
https://github.com/tektoncd/plumbing/blob/ba56b3c3a78bad79177b4730826d780ddcf36319/prow/config.yaml#L2

If you take this link for example https://tekton-releases.appspot.com/build/tekton-prow/logs/tekton-pipeline-publish-images/1250471362649133057/ and swap `logs` with `pr-logs` it works, this is the opposite of the fix ;-)

sbwsg:
Ah I see.

So if the type is "presubmit" then we want to upload to `pr-logs` otherwise we want to upload to `logs`.

PR #330, opened 2020-04-16
Add a job for building PR images 🐉

vdemeester:
# Changes

This adds a new pipeline to build and publish images if any of them
changed in a pull request.

It will tag those images using the pull-request number.

/cc @afrittoli 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/jobs.yaml
+++ b/tekton/ci/jobs.yaml
@@ -92,3 +92,65 @@ spec:
       #!/bin/sh
       set -ex
       yamllint $(params.folders)
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: build-push-kaniko
+spec:
+  params:
+  - name: IMAGE
+    description: Name (reference) of the image to build.
+  - name: DOCKERFILE
+    description: Path to the Dockerfile to build.
+    default: ./Dockerfile
+  - name: CONTEXT
+    description: The build context used by Kaniko.
+    default: ./
+  - name: EXTRA_ARGS
+    default: ""
+  - name: BUILDER_IMAGE
+    description: The image on which builds will run
+    default: gcr.io/kaniko-project/executor:latest
+  resources:
+    inputs:
+    - name: source
+      type: git
+    outputs:
+    - name: endtrigger
+      type: cloudEvent
+  results:
+  - name: IMAGE_DIGEST
+    description: Digest of the image just built.
+
+  steps:
+  - name: build-and-push
+    workingDir: /workspace/source
+    image: $(params.BUILDER_IMAGE)
+    # specifying DOCKER_CONFIG is required to allow kaniko to detect docker credential
+    # https://github.com/tektoncd/pipeline/pull/706
+    env:
+    - name: DOCKER_CONFIG
+      value: /tekton/home/.docker
+    command:
+    - /kaniko/executor
+    - $(params.EXTRA_ARGS)
+    - --dockerfile=$(params.DOCKERFILE)
+    - --context=$(workspaces.source.path)/$(params.CONTEXT)  # The user does not need to care the workspace and the source.
+    - --destination=$(params.IMAGE)
+    - --oci-layout-path=$(workspaces.source.path)/image-digest
+    securityContext:
+      runAsUser: 0
+  - name: write-digest
+    workingDir: $(workspaces.source.path)
+    image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/imagedigestexporter:v0.11.1
+    # output of imagedigestexport [{"name":"image","digest":"sha256:eed29..660"}]
+    command: ["/ko-app/imagedigestexporter"]
+    args:
+    - -images=[{"name":"$(params.IMAGE)","type":"image","url":"$(params.IMAGE)","digest":"","OutputImageDir":"$(workspaces.source.path)/image-digest"}]
+    - -terminationMessagePath=image-digested
+  - name: digest-to-results
+    workingDir: $(workspaces.source.path)
+    image: stedolan/jq
+    script: |
+      cat image-digested | jq '.[0].digest' | tee /tekton/results/IMAGE_DIGEST
--- a/tekton/ci/plumbing-template.yaml
+++ b/tekton/ci/plumbing-template.yaml
@@ -72,10 +72,10 @@ spec:
   - apiVersion: tekton.dev/v1beta1
     kind: PipelineRun
     metadata:
-      name: tekton-random-check-$(uid)
+      name: plumbing-ko-build-$(uid)
       labels:
         prow.k8s.io/build-id: $(params.buildUUID)
-        tekton.dev/check-name: tekton-random-check
+        tekton.dev/check-name: plumbing-ko-build
         tekton.dev/kind: ci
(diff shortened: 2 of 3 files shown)


vdemeester:
Updated with separated pipeline and thus `checks` :angel: 
@afrittoli should we remove the `random` and `noop` check then ? :stuck_out_tongue_closed_eyes:

tekton/ci/jobs.yaml
```diff
@@ -65,3 +65,65 @@ spec:
...
+    # output of imagedigestexport [{"name":"image","digest":"sha256:eed29..660"}]
+    command: ["/ko-app/imagedigestexporter"]
+    args:
+    - -images=[{"name":"$(params.IMAGE)","type":"image","url":"$(params.IMAGE)","digest":"","OutputImageDir":"$(workspaces.source.path)/image-digest"}]
+    - -terminationMessagePath=image-digested
+  - name: digest-to-results
+    workingDir: $(workspaces.source.path)
+    image: stedolan/jq
+    script: |
+      cat image-digested | jq '.[0].digest' | tee /tekton/results/IMAGE_DIGEST
```
afrittoli, line 129: Would passing `--digest-file /tekton/results/IMAGE_DIGEST` to kaniko achieve the same result here?
See https://github.com/GoogleContainerTools/kaniko/blob/master/README.md#--digest-file

tekton/ci/plumbing-hub-build.yaml
```diff
@@ -0,0 +1,78 @@
...
+        value: $(params.checkName)
+  - name: build-hub
+    runAfter: [check-reset]
+    conditions:
+    - conditionRef: "check-git-files-changed"
+      params:
+      - name: gitCloneDepth
+        value: $(params.gitCloneDepth)
+      - name: regex
+        value: "tekton/images/hub/**"
```
afrittoli, line 62: We should make this an input param of the pipeline, so we could have just one pipeline.

tekton/ci/plumbing-template.yaml
```diff
@@ -69,6 +69,321 @@ spec:
       - name: endtrigger
         resourceRef:
           name: github-check-done
+  - apiVersion: tekton.dev/v1beta1
+    kind: PipelineRun
+    metadata:
+      name: plumbing-ko-build-$(uid)
```
afrittoli, line 75: We should be able to use the same pipeline for all images, just we different params.
The check-name label is set here, so it should be fine.

tekton/ci/jobs.yaml
vdemeester: oh this is new :sweat_smile: :angel:

tekton/ci/plumbing-hub-build.yaml
vdemeester: Ahhh one pipeline and multiple pipelineruns, I like it, I'll update :wink:

tekton/ci/plumbing-hub-build.yaml
```diff
@@ -0,0 +1,78 @@
...
+kind: Pipeline
+metadata:
+  name: plumbing-hub-build
+  namespace: tektonci
+spec:
+  params:
+    - name: passorfail
+      description: Should the CI Job 'pass' or 'fail'
+    - name: message
+      description: Message to be logged in the CI Job
```
afrittoli, line 11: We probably don't need these two :)

tekton/ci/plumbing-template.yaml
vdemeester: done :angel:

tekton/ci/tekton-image-build.yaml
```diff
@@ -0,0 +1,81 @@
...
+      description: Number of commits in the change + 1
+    - name: fileFilterRegex
+      description: Names regex to be matched in the list of modified files
+    - name: checkName
+      description: The name of the GitHub check that this pipeline is used for
+    - name: imageName
+      description: The imageName to build
+    - name: imagePrefix
+      description: The prefix for the image reference to build
+      default: gcr.io/tekton-nightly/dogfooding/
```
vdemeester, line 18: Wasn't sure what image prefix to use :angel: 
@afrittoli is the serviceAccount allowed to push there ? (I didn't want to pollute `gcr.o/tekton-release/…` :stuck_out_tongue_closed_eyes: )

tekton/ci/plumbing-template.yaml
```diff
@@ -85,8 +400,6 @@ spec:
       pipelineRef:
         name: tekton-random-check
       params:
-        - name: message
-          value: "BuildUUID#: $(params.buildUUID), Package: $(params.package), Repo: $(params.gitRepository), Revision: $(params.gitRevision)"
```
afrittoli, line 89: Did you mean to remove this from the `tekton-random-check`?

tekton/ci/plumbing-template.yaml
```diff
@@ -69,6 +69,321 @@ spec:
...
+        name: tekton-image-build
+      params:
+        - name: imageName
+          value: ko
+        - name: pullRequestNumber
+          value: $(params.pullRequestNumber)
+        - name: gitCloneDepth
+          value: $(params.gitCloneDepth)
+        - name: fileFilterRegex
+          value: ".*"
```
afrittoli, line 95: shouldn't this be narrower?

tekton/ci/jobs.yaml
afrittoli: so, you prefer to use the imagedigestexporter?

tekton/ci/plumbing-template.yaml
vdemeester: I can :wink:

tekton/ci/plumbing-template.yaml
vdemeester: ah right, we could :angel:

tekton/ci/jobs.yaml
vdemeester: as of this PR yes, but I'll prepare an update in `catalog` and will update this too :angel:

tekton/ci/tekton-image-build.yaml
afrittoli: Good point, the service account used for CI jobs is not allowed to pushed anywhere right now.
`tekton-nightly` is probably better then `tekton-release`, we could add the relevant secrets but you may have to mount them explicitly in the task similar to what we do in the release job.

Perhaps we could use   `gcr.io/tekton-nightly/dogfooding/ci`

tekton/ci/jobs.yaml
afrittoli: NIT: It would be really useful if the last step logged the full URL of the IMAGE for people to easily copy/paste, but that's something we can add later

tekton/ci/tekton-image-build.yaml
```diff
@@ -0,0 +1,81 @@
...
+        value: $(params.fileFilterRegex)
+      resources:
+        inputs:
+        - name: source
+          resource: source
+    taskRef:
+      name: build-push-kaniko
+    params:
+    - name: IMAGE
+      value: $(params.imagePrefix)/$(params.imageName):pr-$(params.pullRequestNumber)
```
afrittoli, line 72: nice :)

vdemeester:
Rebased :angel:

PR #372, opened 2020-05-05
Introduce a simple web requests image for catalog web requests

GijsvanDulmen:
# Changes

Introduce an image for better web requests in some catalog tasks (see https://github.com/tektoncd/catalog/pull/300).

# Submitter Checklist

review them:

for more details._


afrittoli:
Hello @GijsvanDulmen - thank you for your patch.

I'm afraid this is not the right place for this PR. We do maintain container registry images in this repo, but that is for the sole purpose of running Tekton infrastructure itself. We do not host any end-user  image like catalog images.

We went back and forth several time about images in the catalog, and I proposed we shall discuss the topic again during a catalog / hub WG. The position we have for now is that we will not host images in the catalog repo either, as we are not in a position to maintain them right now. This may change in future.

In the mean while I would recommend using an existing image - there should be one with jq and curl - or else to publish one to docker hub. Thank you for your understanding.

afrittoli:
I will close this PR for now.

PR #388, opened 2020-05-13
Fix a small link typo, add the word "webhook" and ci-webhook entry

sbwsg:
# Changes

I went searching for a bit of info on dogfooding secrets and got lost when
I couldnt find the word "webhook". I should've read more closely but
figured I'd add the term here since it's what they're labelled in
GitHub as well.

Also fixes a small link type (goverance.md -> governance.md) and adds
an entry describing the ci-webhook Secret in the tektonci namespace.

# Submitter Checklist

review them:


sbwsg:
huzzah 🎊

PR #389, opened 2020-05-13
Add the most important prow credential! 🤫

bobcatfish:
# Changes

We updated the other github tokens we found in the prow cluster but
actually this is the one prow was using, which we found out the hard way
by deleting the old credentials.

Not actually clear now if the other 2 are important at all. The mystery
continues.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,20 @@ we are trying to [dogfood](#the-dogfooding-cluster) more and more.
 
 Secrets which have been applied to the prow cluster but are not committed here are:
 
-- `GitHub` secrets:
+- `GitHub` personal access tokens:
  - `bot-token-github` in the default namespace
  - `bot-token-github` in the github-admin namespace
+ - `hmac-token` for authenticating GitHub
+ - `oauth-token` which is a GitHub access token for [`tekton-robot`](https://github.com/tekton-robot),
+   used by Prow itself as well as by containers started by Prow via [the Prow config](config.yaml).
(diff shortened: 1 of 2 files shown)


bobcatfish:
/hold

one sec i just found https://github.com/tektoncd/plumbing/tree/master/prow#creating-the-secrets as well

README.md
```diff
@@ -57,6 +57,7 @@ Secrets which have been applied to the prow cluster but are not committed here a
 - `GitHub` secrets:
  - `bot-token-github` in the default namespace
  - `bot-token-github` in the github-admin namespace
+ - `ouath-token` in the default namespace
```
afrittoli, line 60: shall we add the `hmac-token` too? it's the one used on the webhook

sbwsg, line 60: would it be worth mentioning that these are "personal access tokens" specifically?  (Or is this already mentioned elsewhere and I missed it?) I got a bit lost poking around the github UI looking for "secrets", which do exist but are a different concept in Github land apparently.

bobcatfish:
/hold cancel

PTAL!

README.md
```diff
@@ -54,9 +54,21 @@ we are trying to [dogfood](#the-dogfooding-cluster) more and more.
 
 Secrets which have been applied to the prow cluster but are not committed here are:
 
-- `GitHub` secrets:
+- `GitHub` personal access tokens:
  - `bot-token-github` in the default namespace
  - `bot-token-github` in the github-admin namespace
+ - `ouath-token` in the default namespace
```
sbwsg, line 60: nit: ouath

also this is a dupe of `oauth-token` on line 62 now i think?

bobcatfish: 😅  whoops, fixed!

PR #336, opened 2020-04-17
Add a yamllint job

vdemeester:
# Changes

Similar to #330 but for linting yamls

- Adds a yamllint image
- Adds a yamllint nightly build cron
- Adds a yamllint ci job (GitHub check, …)

/cc @afrittoli 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #401, opened 2020-05-26
Add golang lint CI Job

### LINKED ISSUE #378: [feature] Create a Tekton based CI job for go linting
# Expected Behavior
We run the go linter in dedicated CI jobs driven by Tekton

# Actual Behavior
We run the linter as part of the build job in Prow

# Additional Info
This CI job is used by several repositories, so it should be defined in the plumbing repo.
We should use the existing event listener and add a trigger to the list for each repo we want to run the job on. I

This issue can be considered done once we have the job running against one repo (it could be plumbing itself or experimental) and then we can add it to the other repos overtime.

danielhelfand:
Closes #378 

Adding a golang lint CI job to be run against the plumbing repository. Follows a very similar structure to #336 and makes use of `golangci-lint` image. 

# Submitter Checklist

### DIFF
--- a/tekton/ci/jobs.yaml
+++ b/tekton/ci/jobs.yaml
@@ -69,6 +69,55 @@ spec:
 apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
+  name: golang-lint
+  namespace: tektonci
+  description: |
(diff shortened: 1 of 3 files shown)


bobcatfish:
that was a good /meow

PR #408, opened 2020-06-05
Add a Dockerfile that adds a non-root user to alpine/git base

sbwsg:
# Changes

In order to test Pipelines credentials in environments without
root users (i.e. in openshift clusters, or other more locked-down
envs) we need a container image that has a non-root user along
with a tool to exercise the credentials.

This commit adds an alpine-git-nonroot image to our plumbing repo
for use in testing credentials in non-root environments. It uses
alpine/git as its base and then adds a user called nonroot with
UID 1000 to it. This image's first intended use is as a Step in
an example TaskRun that will test creds-init credentials with
non-root securityContext. That PR is ongoing here:
https://github.com/tektoncd/pipeline/pull/2671

# Submitter Checklist

### DIFF
--- a/tekton/ci/plumbing-template.yaml
+++ b/tekton/ci/plumbing-template.yaml
@@ -517,3 +517,48 @@ spec:
       - name: endtrigger
         resourceRef:
           name: github-check-done
+  - apiVersion: tekton.dev/v1beta1
+    kind: PipelineRun
+    metadata:
+      name: plumbing-alpine-git-nonroot-build-$(uid)
+      labels:
+        prow.k8s.io/build-id: $(params.buildUUID)
+        tekton.dev/check-name: plumbing-alpine-git-nonroot-build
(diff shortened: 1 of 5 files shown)


sbwsg:
> You may want to add a CI pipeline (and thus check) for this pipeline 👼 (see tekton/ci folder)

ok, I've added a nightly cronjob for the image, along with an entry in `tekton/ci/plumbing-template.yaml`... is that what I need to do for this or is there something else?  Also... is this process documented anywhere?  I'm just copy/pasting what I see for other images...

vdemeester:
@sbwsg yeah that is perfect I think :wink: We are going to document this better and most likely do a walkthrough/demo on the overall ci/cd on plumbing to help contributors :angel: :crossed_fingers:

afrittoli:
/mario build tekton/images/alpine-git-nonroot alpine-git-nonroot:mario

PR #412, opened 2020-06-09
Use lowercase project name in installation one-liner 🎣

bobcatfish:
# Changes

In tektoncd/pipeline#2786 @ckadner pointed out that the one liner installation line
didn't work because "Pipeline" was capitalized in the path (kind of
surprising that the path is case-sensitive but that's another story!)
It seems that `TEKTON_PROJECT=${p^}` is a bash 5.0 syntax that
capitalizes the word and in the one liner we are using that value, so
instead we will use the uncapitalized value and only captialize in the
spot where we need it.

Closes tektoncd/pipeline#2786

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/resources/release/github_release.yaml
+++ b/tekton/resources/release/github_release.yaml
@@ -50,11 +50,10 @@ spec:
       script: |
         #!/bin/bash
         set -ex
-        p=$(basename $PROJECT)
-        TEKTON_PROJECT=${p^}
+        TEKTON_PROJECT=$(basename $PROJECT)
 
         cat <<EOF | tee $HOME/release.md
-        Tekton $TEKTON_PROJECT release ${VERSION} "$(params.release-name)"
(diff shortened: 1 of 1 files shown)


vdemeester:
@bobcatfish I've updated the description to auto close tektoncd/pipeline#2786 on merge :wink:

tekton/resources/release/github_release.yaml
```diff
@@ -50,11 +50,10 @@ spec:
       script: |
         #!/bin/bash
         set -ex
-        p=$(basename $PROJECT)
-        TEKTON_PROJECT=${p^}
+        TEKTON_PROJECT=$(basename $PROJECT)
```
chmouel, line 53: FYI: If you wanted to do fancy you could have maybe done also : 

```${PROJECT##*/}```

but `basename` works too 😅

PR #436, opened 2020-06-22
Pin the version of gloud that we use

afrittoli:
# Changes

Pin the version of gloud that we use

And since we're changing the image, use the alpine one which is
six times smaller than the debian based one.

# Submitter Checklist

review them:

for more details._

/kind cleanup

### DIFF
--- a/tekton/images/ko-gcloud/Dockerfile
+++ b/tekton/images/ko-gcloud/Dockerfile
@@ -11,7 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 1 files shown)


afrittoli:
> /lgtm
> On _nit_ question: any reason to go on `alpine` ? 😝

It's six times smaller than the debian one. I guess once it's cached it doesn't matter too much, still less code hopefully means fewer issues, vulnerabilities, complexity.

vdemeester:
> It's six times smaller than the debian one. I guess once it's cached it doesn't matter too much, still less code hopefully means fewer issues, vulnerabilities, complexity.

On the downside, it's using `musl` which means it's "less" battle tested in a way… In any case, it's a nit :wink: 
/approve

PR #443, opened 2020-06-24
Update Dashboard Robocat deployment cronjob to match pipeline and triggers

sbwsg:
# Changes

The dashboard robocat deployment cronjob was using extra values that don't
appear in the pipeline or triggers cronjobs.

This PR makes the cronjobs match and hopefully will allow the dashboard
nightly deployments in robocat to resume.

# Submitter Checklist

review them:

### DIFF
--- a/tekton/cronjobs/robocat-dashboard-deploy-latest-cron/cronjob.yaml
+++ b/tekton/cronjobs/robocat-dashboard-deploy-latest-cron/cronjob.yaml
@@ -17,12 +17,13 @@ spec:
               - name: TEKTON_PROJECT
                 value: "dashboard"
               - name: TEKTON_VERSION
-                value: "beta"
+                value: "latest"
               - name: TEKTON_CLUSTER
                 value: "robocat"
               - name: RELEASE_FILE
+                # We deploy the read-only dashboard since it will be exposed to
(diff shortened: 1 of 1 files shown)


tekton/cronjobs/robocat-dashboard-deploy-latest-cron/cronjob.yaml
```diff
@@ -17,12 +17,8 @@ spec:
               - name: TEKTON_PROJECT
                 value: "dashboard"
               - name: TEKTON_VERSION
-                value: "beta"
+                value: "latest"
               - name: TEKTON_CLUSTER
                 value: "robocat"
-              - name: RELEASE_FILE
-                value: "tekton-dashboard-release-readonly.yaml"
```
afrittoli, line 24: We should not change the release filename. 
Without this we will deploy the non-readonly dashboard, which would give the world access to the robocat cluster.

sbwsg: 👍 Thanks for clarifying that, I've put back the original value and added a comment.

AlanGreene:
Looks good, thanks 🤞 

/lgtm

PR #456, opened 2020-07-03
Update releases task to use workspaces…

vdemeester:
# Changes

… instead of PipelineResources. This is required for some changes in
the tektoncd/pipeline release tasks.

Linked to https://github.com/tektoncd/pipeline/pull/2897

/kind misc
/cc @afrittoli 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/resources/release/github_release.yaml
+++ b/tekton/resources/release/github_release.yaml
@@ -25,10 +25,12 @@ spec:
     description: Release number and git tag to be applied (e.g. 0.888.1, with 'v')
   - name: previous-release-tag
     description: Previous release number - for author and PR list calculation
+  - name: revision
+    description: The git commit hash that will be tagged with the release
+  workspaces:
+  - name: source
   resources:
     inputs:
-      - name: source
-        type: git
       - name: release-bucket
         type: storage
   stepTemplate:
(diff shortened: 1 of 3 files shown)


tekton/resources/release/github_release.yaml
```diff
@@ -74,16 +76,16 @@ spec:
...
       image: gcr.io/tekton-releases/dogfooding/hub
-      workingdir: $(resources.inputs.source.path)
+      workingdir: $(workspaces.source.path)
       script: |
         #!/usr/bin/env bash
         set -ex
 
         # Restore full git history
-        git fetch --unshallow
+        git fetch --unshallow || true
```
afrittoli, line 85: If this fails we should fail.

vdemeester: right, *kinda* :stuck_out_tongue: 
The *main* problem is that, if the workspace *was* fully fetch (by a previous task, …), this fails because it is already *unshallowed*. I didn't see any way to make it fail only if the fetch fail and not if the repo was already *unshallowed*.. :sweat_smile:

vdemeester: My take on adding the `|| true` is that if the fetch fail for any other reason that the repo is already unshallowed, the following commands (`git log` or `hub …`) will fail :sweat_smile:

PR #470, opened 2020-07-08
Deploy dogfooding nightly

afrittoli:
# Changes

Add a tekton-cd account that is admin on default, tektonci and
mario namespaces (it can control all resource types that are
included in the deploy in those namespaces).
Add a pipeline resource associated to that service account.

Add a cron job that deploys resources from the tekton folder in
the plumbing repo to the dogfooding cluster, using the tektoncd
account. Runs at 10:30pm UTC daily.

Add a cron job that deploys cronjobs from the tekton folder in
the plumbing repo to the dogfooding cluster, using the tektoncd
account. Runs at 9:30pm UTC daily.

Since kubectl includes version v2.0.3 of kustomize, which still
uses bases instead of resources, add kustomize v3 to the kubectl
image, and change the folder trigger template to use kustomize
build for overlays instead of kubectl -k.

Improve the folder deployment template:
Fix the replace mode by adding create + replace, and switch the
default deploy mode to replace. Check for failures in create,
if any failure other than "already exists" is found, fail.
Never force replace, as that deletes and recrates, which can
cause all sort of issues when applied to resources like
service accounts or ingresses.

Improve efficiency by checking if any diff exists, and skip
deployment if not. This is possible now that Tekton resources
support diff.

# Submitter Checklist

review them:

for more details._

/kind feature

### DIFF
--- a/tekton/cronjobs/bases/folder/trigger-resource-cd.yaml
+++ b/tekton/cronjobs/bases/folder/trigger-resource-cd.yaml
@@ -69,7 +69,7 @@ spec:
               - name: CLUSTER_RESOURCE
                 value: "not-a-real-cluster"
               - name: FOLDER_METHOD
-                value: "apply"
+                value: "replace"
               - name: FOLDER_OVERLAY
                 value: "false"
               - name: FOLDER_PATH
--- a/tekton/resources/cd/folder-template.yaml
+++ b/tekton/resources/cd/folder-template.yaml
@@ -79,16 +79,38 @@ spec:
             NAMESPACE_PARAM="-n $(params.namespace)"
             [[ "$(params.namespace)" == "" ]] && NAMESPACE_PARAM=""
 
-            # Run the deployment
+            # Handle overlays
+            TARGET=$(resources.inputs.source.path)/$(params.folderPath)
             if [[ "$(params.isOverlay)" == "true" ]]; then
-              kubectl "$(params.deployMethod)" \
-                $NAMESPACE_PARAM -f \
-                $(resources.inputs.source.path)/$(params.folderPath)
-            else
+              TARGET=target.yaml
               kustomize build \
-                $(resources.inputs.source.path)/$(params.folderPath) | \
-                kubectl "$(params.deployMethod)" $NAMESPACE_PARAM -f -
+                $(resources.inputs.source.path)/$(params.folderPath) > $TARGET
             fi
+
+            # Check if there is any diff
+            DIFF=diff.txt
+            kubectl diff $NAMESPACE_PARAM -f $TARGET | tee $DIFF
+
+            # If there is no diff, we don't need to update
+            if [ ! -s ${DIFF?} ]; then
+              echo "No change detected, nothing to be done."
+              exit 0
+            fi
+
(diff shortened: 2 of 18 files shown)


tekton/cronjobs/bases/folder/trigger-resource-cd.yaml
```diff
@@ -69,7 +69,7 @@ spec:
               - name: CLUSTER_RESOURCE
                 value: "not-a-real-cluster"
               - name: FOLDER_METHOD
-                value: "apply"
+                value: "replace"
```
vdemeester, line 72: :heart_eyes_cat:

tekton/resources/cd/folder-template.yaml
```diff
@@ -75,18 +75,42 @@ spec:
             #!/bin/sh
             set -ex
```
vdemeester, line 76: We *may* want to start migrating thoses to scripts in the repo ? (easier to change/debug, run independently)

tekton/resources/cd/folder-template.yaml
```diff
@@ -75,18 +75,42 @@ spec:
...
+
+            # If there is no diff, we don't need to update
+            if [ ! -s ${DIFF?} ]; then
+              echo "No change detected, nothing to be done."
+              exit 0
+            fi
+
+            # When deploying with replace, we need to do a create first,
+            # to ensure new resources are created
+            CREATE_OUTPUT=create.txt
```
vdemeester, line 102: Do we have a way to detect resources that were created before (by this), and not created anymore (aka we should remove them) ?

afrittoli: I added a "managed-by" annotation for now, but we could make that a label. That way we could search for all resources managed by tekton and delete those that don't show up in the manifests we're deploying.
We could even add more granularity as we have multiple cronjobs doing deployment, so each should care about its own resources only.

tekton/resources/cd/folder-template.yaml
afrittoli: Yes, the are also downsides to doing that though.
The task will need a git repo to get the script from (I guess it could fetch it via Https too?) and the task and script version will be split. If we keep the task in here it's fine, but if we add it to the catalog, how would that work? Can we host scripts in the catalogue? Is there perhaps something we should add to the script mode? I.e. the ability to specify a script URL?

afrittoli: Is this something we can sort out *after* this PR?

tekton/resources/cd/folder-template.yaml
afrittoli: can we add support for deletion as a follow-up?

tekton/resources/cd/folder-template.yaml
vdemeester: @afrittoli of course :+1: it was a general "wondering" :stuck_out_tongue_closed_eyes:

tekton/resources/cd/folder-template.yaml
vdemeester: yes :+1:

PR #480, opened 2020-07-16
test-runner: do not use -u on go get…

vdemeester:
# Changes

Using `go get -u`, Go will try to update dependencies when getting the
package. This is bad with go modules as there is no guarantee the new
dependencies still work (especially when the dependency is 0.x). As we
are only installing the binaries, we need reproductible install, and…
using `go get -u` is the opposite of that 😅

*Also bump golangci-lint to 1.28.0*.

# Submitter Checklist

review them:

for more details._


chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

PR #498, opened 2020-08-02
Handle the issue of labels with quotes within description

afrittoli:
# Changes

Labels are extracted from the JSON payload of the github
event, so they must be a valid JSON fragment. The problem
of labels with single quotes in the description begins when
we assign that JSON fragment to a variable so we can later
parse it with json.loads.

The solution to that is to use the triple-quotes syntax
to define the variable that holds the json.

# Submitter Checklist

review them:

for more details._

/kind bug
/cc @vdemeester

### DIFF
--- a/tekton/ci/checks/check-pr-has-kind-label.yaml
+++ b/tekton/ci/checks/check-pr-has-kind-label.yaml
@@ -82,7 +82,8 @@ spec:
             import yaml
             import sys
 
-            prLabels = json.loads('$(params.labels)')
+            prLabelsText = """$(params.labels)"""
+            prLabels = json.loads(prLabelsText)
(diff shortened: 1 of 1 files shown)


tekton/ci/checks/check-pr-has-kind-label.yaml
```diff
@@ -82,7 +82,8 @@ spec:
             import yaml
             import sys
 
-            prLabels = json.loads('$(params.labels)')
+            prLabelsText = """$(params.labels)"""
+            prLabels = json.loads(prLabelsText)
             labelNames = list(map(lambda e: e["name"], prLabels))
```
chmouel, line 87: nit, why are you not using just list comprehension instead of functional programming heuristics, like map, lambdas and filters :)

Issue #516, opened 2020-08-14
Triggers and Dashboard Nightly failing on git resource


dibyom:
# Expected Behavior
Nightlies should pass

# Actual Behavior

The git init step fails:

```
{"level":"error","ts":1597372222.9066162,"caller":"git/git.go:47","msg":"Error running git [fetch --recurse-submodules=yes --depth=1 origin --update-head-ok --force f39feca70117455dbd765cdb9a3c8fbcad134977]: exit status 128\nfatal: no path specified; see 'git help pull' for valid url syntax\n","stacktrace":"github.com/tektoncd/pipeline/pkg/git.run\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:47\ngithub.com/tektoncd/pipeline/pkg/git.Fetch\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:137\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:52\nruntime.main\n\truntime/proc.go:203"}
{"level":"fatal","ts":1597372222.9067645,"caller":"git-init/main.go:53","msg":"Error fetching git repository: failed to fetch [f39feca70117455dbd765cdb9a3c8fbcad134977]: exit status 128","stacktrace":"main.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:53\nruntime.main\n\truntime/proc.go:203"}
```

# Additional Info
tkn version:

```
Pipeline version: v0.15.2
Triggers version: v0.7.0
```

https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/triggers-release-nightly-28m2c?pipelineTask=unit-tests&step=git-source-source-mshdx

afrittoli:
I wonder if this is somehow related to the changes in https://github.com/tektoncd/plumbing/pull/514

dibyom:
When  I tried to re-run a CronJob, I did see the Trigger failing when it tried to create a PipelineResource that already existed:

```
[...]

PR #517, opened 2020-08-14
Use embedded storage resources for nightly releases

dibyom:
# Changes

The nightly releases were previously creating a new storage resource for each
run.  This was changed in #514 to re-use the same per project bucket for each
run. However, we still had the PipelineResource in the nightly TriggerTemplate
which means Triggers will try to create the same PipelineResource on each run
-- and in the process fail because the resource already existed.

This commit fixes the error by changing the nightly pipelinerun templates to
use an embedded storage resource.

Found while investigating #516

/kind bug

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #519, opened 2020-08-14
Use embedded storage resources for nightly releases

dibyom:
# Changes

The nightly releases were previously creating a new storage resource for each
run.  This was changed in #514 to re-use the same per project bucket for each
run. However, we still had the PipelineResource in the nightly TriggerTemplate
which means Triggers will try to create the same PipelineResource on each run
-- and in the process fail because the resource already existed.

This commit fixes the error by changing the nightly pipelinerun templates to
use an embedded storage resource.

Found while investigating #516

# Submitter Checklist

review them:

for more details._


dibyom:
/assign @afrittoli 

The old one #517 got closed somehow....

PR #520, opened 2020-08-17
prow: fix query for stale, rotten and close issues

vdemeester:
# Changes

As of today, an issue is close after 30 days (without necessarly been
rotten) because it does filter them out. Same for lifecycle/rotten, it
does not do 30 + 90 days, but only 90 days.

It seemed there was way too much issue being closed :sweat_smile: 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

/cc @jerop @afrittoli @bobcatfish 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1684,7 +1684,7 @@ periodics:
       - |-
         --query=org:tektoncd
         -label:lifecycle/frozen
-        -label:lifecycle/stale
+        label:lifecycle/stale
         -label:lifecycle/rotten
       - --updated=720h
       - --token=/etc/github-token/oauth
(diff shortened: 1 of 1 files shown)


bobcatfish:
I don't understand the query syntax but thanks for fixing this anyway @vdemeester ! :D

vdemeester:
> I don't understand the query syntax but thanks for fixing this anyway @vdemeester ! :D

`-label:lifecycle/frozen` mean "filter issues/prs with label `lifecycle/frozen`" from the query. `label:lifecycle/rotten` (with the dash `-`) means filter those in (aka this will return all issues with that label). The filter was the same for `stale`, `rotten` and `close`, meaning we would apply the 2 at the same time (with the same time base), hence issues being closed after only 30 days of inactivity. Now, we do apply the `rotten` part only on the issues that already have the `lifecycle/stale` label. And *as* adding the comment count as an activity, the 30 days + 90 days are respected :wink: :stuck_out_tongue_closed_eyes:

PR #522, opened 2020-08-17
Add git url to templates for nightlies

sbwsg:
# Changes

The url for the git resources for each of the nightlies is
missing from the trigger template. The nightlies are
currently failing with errors like this:

```
Error running git [fetch --recurse-submodules=yes --depth=1 origin
--update-head-ok --force f39feca70117455dbd765cdb9a3c8fbcad134977]: exit
status 128\nfatal: no path specified; see 'git help pull' for valid url
syntax
```

This commit adds the url to the git resources from the
base trigger template's parameters.

# Submitter Checklist


dibyom:
Ahh, good catch. 
I think this fixes https://github.com/tektoncd/plumbing/issues/516

PR #528, opened 2020-08-18
Update TriggerTemplate overlays to use different path

### LINKED ISSUE #526: Nightly resources don't appear to be building quite right
# Expected Behavior

`kustomize build`ing the nightly release yamls works as expected.

# Actual Behavior

At the moment it looks like trying to build the yamls results in invalid TriggerTemplates for pipelines / dashboard / triggers. The `params` field from the `base/template.yaml` are not appearing in the final output.

# Steps to Reproduce the Problem

1. cd into the plumbing repo root directory
2. `kustomize build ./tekton/resources/nightly-release > ./nightly-resources.yaml`
3. Take a look at the generated `TriggerTemplate`s for pipelines, dashboard, triggers
4. The `params` field of the TriggerTemplate is missing.

sbwsg:
# Changes

Fixes #526 

Problem statement: At the moment our nightly TriggerTemplates are
built without a crucial `params` field that is defined in the base TriggerTemplate.
This means our nightlies aren't building - the params include information like
the correct git repo to fetch.

The current path in the nightly TriggerTemplate overlays is
"/spec". Using this path causes the spec field of the base
template to be completely replaced. The spec field of the base
includes a "params" field that we want to keep in each built
TriggerTemplate. Replacing "/spec" means we lose the "params" field.

This commit changes the path to "/spec/resourcetemplates" and
updates the value to be an array entry. The result is that the
final built document includes both the base template's params
field as well as the overlay's resourcetemplates entry.

HOWEVER, this effectively rolls back a change made in 46137f0
and I'm not entirely sure that this is therefore a good idea.
I might be re-introducing a broken behaviour from before that
commit was made. Therefore I'm creating this PR to seek input
from others on whether this is incorrect since my knowledge of
JSON Patching and of these templates is somewhat limited.

# Submitter Checklist

review them:

/kind bug

### DIFF
--- a/tekton/resources/nightly-release/overlays/dashboard/template.yaml
+++ b/tekton/resources/nightly-release/overlays/dashboard/template.yaml
@@ -1,7 +1,6 @@
 - op: add
-  path: "/spec"
+  path: /spec/resourcetemplates
   value:
-    resourcetemplates:
     - apiVersion: tekton.dev/v1beta1
       kind: PipelineRun
       metadata:
--- a/tekton/resources/nightly-release/overlays/pipeline/template.yaml
+++ b/tekton/resources/nightly-release/overlays/pipeline/template.yaml
@@ -1,7 +1,6 @@
 - op: add
-  path: "/spec"
+  path: /spec/resourcetemplates
   value:
-    resourcetemplates:
     - apiVersion: tekton.dev/v1beta1
       kind: PipelineRun
       metadata:
--- a/tekton/resources/nightly-release/overlays/triggers/template.yaml
+++ b/tekton/resources/nightly-release/overlays/triggers/template.yaml
@@ -1,7 +1,6 @@
 - op: add
-  path: "/spec"
+  path: /spec/resourcetemplates
   value:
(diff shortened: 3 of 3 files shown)


dibyom:
I think this PR does the right thing. The issue with 46137f0 was that we were adding the resourcetemplates to /spec/resourcetemplates/- (i.e. adding resource template to the already existing one in the base). This one seems to do the right thing in my local setup

sbwsg:
This PR is applied in dogfooding and the TriggerTemplate appears to be working again:

http://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/triggers-release-nightly-pmz5h
http://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/dashboard-release-nightly-vxfv7
http://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/pipeline-release-nightly-7lrlm

I think we are good to merge!

dibyom:
/approve

Thanks @sbwsg

PR #532, opened 2020-08-19
Reduce the chance of race condition on github update

afrittoli:
# Changes

It may happen that the "start" event is processed after
the "complete" one. To handle this case, we can check
the check status before updating, and skip the update
if we see that a "complete" was already processed.

This does not fully eliminate the possibility of race,
as the two update jobs may still run in parallel,
but it reduces the chances considerably.

# Submitter Checklist

review them:

for more details._

/kind bug

### DIFF
--- a/tekton/resources/ci/github-template.yaml
+++ b/tekton/resources/ci/github-template.yaml
@@ -62,6 +62,7 @@ spec:
             #!/usr/bin/env python
             import json
             import random
+            import sys
 
             check_status = "$(tt.params.checkStatus)"
             if "$(tt.params.parentPipelineRunName)" == "":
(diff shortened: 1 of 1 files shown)


tekton/resources/ci/github-template.yaml
```diff
@@ -77,6 +78,19 @@ spec:
 
             # we may need to URLEncode the check name
             status_path = "$(resources.outputs.pr.path)/status/$(tt.params.checkName).json"
+            with open(status_path, "r") as status:
+              old_status = json.loads(status)
+
+            # Detect race condition. It may be that the "start" event is processed after
+            # the "complete" one. This does not eliminate the possibility of race as the two
+            # update jobs may still run in parallel, but it reduces the chances considerably.
+            if (status_body['State'] in ['success', 'failure', 'error'] and
```
vdemeester, line 87: Doesn't this prevent from going from `success` or `failure` to `pending` (in case of an update on the PR) ?

afrittoli: It shouldn't because the target URL includes the PR name, which is unique per run, so it will be different after a PR sync (or event after a `/test`

vdemeester: oh nice :+1:

vdemeester (approved):
/lgtm

vdemeester (commented):
/lgtm

vdemeester (commented):
/lgtm

PR #562, opened 2020-09-08
Catlin: Validate images used in task steps are tagged

SM43:
Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/catlin/go.mod
+++ b/catlin/go.mod
@@ -3,6 +3,7 @@ module github.com/tektoncd/plumbing/catlin
 go 1.14
 
 require (
+	github.com/google/go-containerregistry v0.1.1
 	github.com/spf13/cobra v1.0.0
 	github.com/tektoncd/pipeline v0.15.2
 	go.uber.org/zap v1.15.0
--- a/catlin/pkg/cmd/validate/validate.go
+++ b/catlin/pkg/cmd/validate/validate.go
@@ -17,8 +17,10 @@ package validate
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+
 	"github.com/tektoncd/plumbing/catlin/pkg/app"
 	"github.com/tektoncd/plumbing/catlin/pkg/parser"
 	"github.com/tektoncd/plumbing/catlin/pkg/validator"
@@ -74,6 +76,7 @@ func validate(cli app.CLI, path string) error {
 	validators := []validator.Validator{
 		validator.NewPathValidator(cli, res, path),
 		validator.NewContentValidator(cli, res),
+		validator.ForKind(res),
 	}
 
 	result := validator.Result{}
@@ -84,10 +87,18 @@ func validate(cli app.CLI, path string) error {
 	// print result
 	out := cli.Stream().Out
 	for _, v := range result.Lints {
-		if v.Kind == validator.Error {
+		switch v.Kind {
+		case validator.Error:
 			fmt.Fprintf(out, "ERROR: %s\n", v.Message)
-		} else {
+		case validator.Warning:
+			fmt.Fprintf(out, "WARN : %s\n", v.Message)
+		case validator.Recommendation:
+			fmt.Fprintf(out, "HINT : %s\n", v.Message)
+		case validator.Info:
 			fmt.Fprintf(out, "INFO : %s\n", v.Message)
+		default:
+			level := strings.ToUpper(string(v.Kind))
+			fmt.Fprintf(out, "%s : %s\n", level, v.Message)
 		}
 	}
 	if result.Errors != 0 {
--- a/catlin/pkg/validator/task_validator.go
+++ b/catlin/pkg/validator/task_validator.go
@@ -1,0 +1,122 @@
+// Copyright © 2020 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validator
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+	"github.com/tektoncd/plumbing/catlin/pkg/parser"
+)
+
+const (
+	parameterSubstitution = `[_a-zA-Z][_a-zA-Z0-9.-]*(\[\*\])?`
+	braceMatchingRegex    = "(\\$(\\(%s.(?P<var>%s)\\)))"
+)
+
+type taskValidator struct {
+	res *parser.Resource
+}
+
+var _ Validator = (*taskValidator)(nil)
+
+func NewTaskValidator(r *parser.Resource) *taskValidator {
+	return &taskValidator{res: r}
+}
+
+func (t *taskValidator) Validate() Result {
+
+	result := Result{}
+
+	res, err := t.res.ToType()
+	if err != nil {
+		result.Error("Failed to decode to a Task - %s", err)
+		return result
+	}
+
+	task := res.(*v1beta1.Task)
+	for _, step := range task.Spec.Steps {
+		result.Append(t.validateStep(step))
+
+	}
+	return result
+}
+
+func (t *taskValidator) validateStep(s v1beta1.Step) Result {
+
+	result := Result{}
+	step := s.Name
+	img := s.Image
+
+	if _, usesVars := extractExpressionFromString(img, ""); usesVars {
+		result.Warn("Step %q uses image %q that contains variables; skipping validation", step, img)
+		return result
+	}
+
+	if !strings.Contains(img, "/") || !isValidRegistry(img) {
+		result.Warn("Step %q uses image %q; consider using a fully qualified name - e.g. docker.io/library/ubuntu:1.0", step, img)
+	}
+
+	if strings.Contains(img, "@sha256") {
+		rep, err := name.NewDigest(img, name.WeakValidation)
+		if err != nil {
+			result.Error("Step %q uses image %q with an invalid digest. Error: %s", step, img, err)
+			return result
+		}
+
+		if tagHasDigest(rep.String()) {
+			result.Warn("Step %q uses image %q; consider using digest over tags as tags are mutable", step, img)
+		}
+
+		return result
+	}
+
+	ref, err := name.NewTag(img, name.WeakValidation)
+	if err != nil {
+		result.Error("Step %q uses image %q with an invalid tag. Error: %s", step, img, err)
+		return result
+	}
+
+	if strings.Contains(ref.Identifier(), "latest") {
+		result.Error("Step %q uses image %q which must be tagged with a specific version", step, img)
+	}
+
+	return result
+}
+
+// copied from tektoncd/pipeline
+func extractExpressionFromString(s, prefix string) (string, bool) {
+	pattern := fmt.Sprintf(braceMatchingRegex, prefix, parameterSubstitution)
+	re := regexp.MustCompile(pattern)
+	match := re.FindStringSubmatch(s)
+	if match == nil {
+		return "", false
+	}
+	return match[0], true
+}
+
+func isValidRegistry(img string) bool {
+	repo := strings.Split(img, "/")[0]
+	return strings.Contains(repo, ".")
+}
+
+func tagHasDigest(img string) bool {
+	withOutDigest := strings.Split(img, "@sha256")[0]
+	return strings.Contains(withOutDigest, ":")
+}
--- a/catlin/pkg/validator/task_validator_test.go
+++ b/catlin/pkg/validator/task_validator_test.go
@@ -1,0 +1,204 @@
+// Copyright © 2020 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validator
+
+import (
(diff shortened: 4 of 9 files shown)


catlin/pkg/validator/kind_validator.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func NewKindValidator(cli app.CLI, r *parser.Resource) Validator {
+
+	switch r.Kind {
+	case "Task":
+		return &TaskValidator{res: r, log: cli.Logger().Sugar()}
+	}
+
+	return nil
```
sthaha, line 29: this should return an error `UnknownKindError` instead of nil which can be checked by the caller before adding to the list of `validators`

catlin/pkg/validator/kind_validator.go
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+package validator
+
+import (
+	"github.com/tektoncd/plumbing/catlin/pkg/app"
+	"github.com/tektoncd/plumbing/catlin/pkg/parser"
+)
+
+func NewKindValidator(cli app.CLI, r *parser.Resource) Validator {
```
sthaha, line 22: How about `ValidatorForKind` that returns `Validator, error` ?

catlin/pkg/validator/task_validator.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+var _ Validator = (*TaskValidator)(nil)
+
+func (t *TaskValidator) Validate() Result {
+
+	result := Result{}
+
+	var task *v1beta1.Task
+	str, _ := json.Marshal(t.res.Unstructured.Object)
```
sthaha, line 39: I think it may be better to add a function to the `ParsedResource` to convert to the right type. 
```go
var task v1beta.Task
err := r.ToType(&task)
```
and then 

```
for _, step := ... 
```

This makes it quite easy to add  support for other types as well. Please refer to the code in parse to properly unmarshall to a typed object.

SM43: Thank you :)

catlin/pkg/validator/task_validator.go
```diff
@@ -0,0 +1,52 @@
...
+		result.Error("failed to decode task - %s", err)
+		return result
+	}
+
+	for _, step := range res.(*v1beta1.Task).Spec.Steps {
+		img := strings.FieldsFunc(step.Image, func(c rune) bool { return c == ':' })
+
+		if len(img) == 1 || img[1] == "latest" {
+			result.Error("Task image (%s) must be tagged with a specific version", step.Image)
+		}
```
vdemeester, line 48: We should use a package that parse image reference, like `go-containerregistry` `pkg/name` package or docker's packages. This doesn't take digests into account for example (although this code would work)

catlin/pkg/cmd/validate/validate.go
```diff
@@ -76,6 +76,11 @@ func validate(cli app.CLI, path string) error {
 		validator.NewContentValidator(cli, res),
 	}
 
+	kindValidator, err := validator.ValidatorForKind(cli, res)
```
sthaha, line 79: nit: 
```go
if v, err := validator.ForKind(cli, res.Kind); err == nil {
  validators = append(validators, v)
}
```

On a second thought, it seems like an anti-pattern to ignore error. How about a `NoopValidator` if kind has no specific validator and `err` will a `not nil` if there was a failure to create validator at which point the CLI should exit with that error? 
So .. the code would look like 

```go 
v, err := validator.ForKind(res.Kind)
if err != nil {
  return err
}
validators = append(validators, v)

```

catlin/pkg/validator/task_validator.go
sthaha: Good point! I didn't think of it at all! 🤗

SM43: `digest` would not be mandatory right? the image should be tagged with a version and not `latest`. Is that correct?

vdemeester: @SM43 that's correct. Using a digest, `foo@sha256:abc…` would work too. I think, the current code "works" with digests, but *by change*. Although it's not gonna change in the long-term future, using already existing like to parse/extra part of the image reference is a long-term viable thing to do (allow to validate more accurately the image reference, etc.)

catlin/pkg/cmd/validate/validate.go
SM43: When do we want `ForKind` to return an error? we are just checking the kind and returning a validator if exists otherwise we return `NoopValidator`. we aren't doing anything on the resource inside as it is already parsed. so, not sure when we should return an error :thinking:
If res is of  `UnknownKind` the parser would fail and return from there.

catlin/pkg/validator/testdata.go
```diff
@@ -0,0 +1,83 @@
...
+    tekton.dev/displayName: My Example Task
+spec:
+  description: |-
+    A summary of the resource
+
+    A para about this valid task
+
+  steps:
+  - name: hello
+    image: ubuntu
```
sthaha, line 50: @vdemeester @chmouel is this valid? shouldn't we insist on a tag ? 
cc: @SM43

catlin/pkg/validator/testdata.go
```diff
@@ -0,0 +1,83 @@
...
+  - name: foo-bar
+    image: abc.io/fedora:1.0@sha256:deadb33fdeadb33fdeadb33fdeadb33fdeadb33fdeadb33fdeadb33fdeadb33f
+`
+
+	taskWithInvalidImageRef = `
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: valid
```
sthaha, line 35: this need to be called `invalid` 😂

catlin/pkg/validator/testdata.go
sthaha: this will fail validation :D 
I went by the name of the task - `valid`

vdemeester: This would be invalid indeed.

catlin/pkg/validator/task_validator_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	res, err := parser.Parse()
+	assert.NilError(t, err)
+
+	v := ForKind(res)
+	result := v.Validate()
+
+	assert.Equal(t, 3, result.Errors)
+	assert.Equal(t, 3, len(result.Lints))
+	assert.Equal(t, `Step "hello" has an invalid image: could not parse reference: ubuntu`, result.Lints[0].Message)
```
vdemeester, line 54: wait, it should be the same error as `abc.io/fedora:latest` here. `ubuntu` is valid (because it's an "official" image), as it would become `docker.io/library/ubuntu` :upside_down_face:

SM43: `docker.io/library/ubuntu` would be valid but `ubuntu` invalid as per current implementation. Is that okay?

sthaha: @vdemeester `ubunbu` without a tag or a sha is valid?

sthaha: go-containerreg fails to parse reference to `ubuntu` - right? isn't that why we are getting this error?

SM43: `go-containerregistry/pkg/name.ParseReference` fails to validate the image ref as it does not have registry appended `docker.io`. Is there a way to which know is an official image?

sthaha: IIUC, https://github.com/google/go-containerregistry/blob/093adf2740a9617fbb6e6fe81ec37b00d3eb1b43/pkg/name/digest_test.go would fail validation of `ubuntu`

vdemeester: oh :joy_cat: so yeah, `ubuntu` should be valid. It is *always* better to use the full ref (`docker.io/library/ubuntu`) but `ubuntu` should be valid. An *official* image is like `ubuntu`, no username/group or host on top of it. `traefik`, `debian`, `fedora`, `mysql`, … all those are "official" images

vdemeester: We could *warn* in case of those, recommending using the full ref..

vdemeester: I am amazed `go-containerregistry` fails on those though…

vdemeester: Oh, I guess it depends on the `opt.strict` option thouhg…

SM43: Yeah. If we use `WeakValidation` then it doesn't fail for ubuntu but also for any image name for eg. `api` or `hub`
How do we differentiate between official and non-official images or should we warn for both?

vdemeester: So, I am fine using weak validation and have a `hub` image being seen as valid, even though it doesn't exists — as `foo.com/bar/baz:mytag` is gonna be seen as valid even though it doesn't exists.

And, yeah, I would warn for any image that is just *one part*, aka `ubuntu`, `api`, … being "official" or not, with a message "prefer using a fully qualified reference, like `library/ubuntu` or `docker.io/library/ubuntu` instead".

SM43: okay then Is `tag` required for all images?

vdemeester: a tag is required for all images and should be an error. images like `hub` or `ubuntu` should have a warn telling to use a fully qualified reference, like `library/hub`, `docker.io/library/ubuntu`.

catlin/pkg/validator/task_validator_test.go
```diff
@@ -0,0 +1,80 @@
...
+    image: api
+  - name: missing-repo
+    image: api:1.0
+  - name: no-tags
+    image: quay.io/api
+  - name: valid
+    image: registry.io/api:1.0
+`
+
+func TestTaskValidator_InvalidImageRef(t *testing.T) {
```
sthaha, line 48: @vdemeester @chmouel Could you please `lgtm` if this test covers all/most cases? 
To me this seems to cover all cases but will wait for you to do the honours :)

vdemeester: imho, this is too involved :sweat_smile: 
- I would rather not do any whitelist at all, and warn in all the case where a *one-word* image is used (be it `ubuntu`, `alpine` or `foobar`.
- according to the spec, nothing should make `quay.io/api` invalid. We **know** it is, because we know quay, but a program can't really make that assumption (and if it does, it feels weird). An exemple with https://r.j3ss.co/, the image references there are `r.j3ss.co/clisp:latest` and should be considered as valid.

catlin/pkg/validator/task_validator_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// whitelised image produces warning
+	assert.Equal(t, Warning, lints[0].Kind)
+	assert.Equal(t, `Step "whitelisted" uses image "ubuntu"; consider using a fully qualified name - e.g. docker.io/library/ubuntu`, lints[0].Message)
+
+	// image with variables produces warning
+	assert.Equal(t, Warning, lints[1].Kind)
+	assert.Equal(t, `Step "variables" uses image "ubuntu:$(params.img_ver)" that contains variables; skipping validation`, lints[1].Message)
+
+	// image with variables produces warning
```
vdemeester, line 71: that comment is not right as it is errors :stuck_out_tongue_closed_eyes:

vdemeester:
Let me add some more validation examples
- `ubuntu` should be 1 warning (use full ref, aka docker.io/library/ubuntu) and 1 error (use a tag, different than latest)
- `ubuntu:latest` should be 1 warning (use full ref, aka docker.io/library/ubuntu) and 1 error (use a tag, different than latest)
- `foo` should be 1 warning (use full ref, aka docker.io/library/foo) and 1 error (use a tag, different than latest)
- `foo:bar` should be 1 warning (use full ref, aka docker.io/library/foo)
- `docker.io/foo` should be 1 error (use a tag, different than latest)
- `docker.io/foo:bar` should pass
- `r.j3ss.co/clisp` should be 1 error (use a tag, different than latest)
- `r.j3ss.co/clisp:latest` should be 1 error (use a tag, different than latest)
- `docker.io/vdemeester/foo` should be 1 error (use a tag, different than latest)
- `docker.io/vdemeester/foo:baz` should pass
- `quay.io/foo` should be 1 error (use a tag, different than latest)
- `quay.io/foo:bar` should pass
- `gcr.io/foo/bar/baz` should be 1 error (use a tag, different than latest)
- `gcr.io/foo/bar/baz:bar` should pass

There is more too
- `localhost:5000/foo:bar` should **probably** fail
- `172.16.3.4:3000/foo/bar:baz` should pass

catlin/pkg/validator/task_validator_test.go
chmouel: 👍 on @vdemeester I dont think we should make it over complicated.

catlin/pkg/validator/validate_test.go
```diff
@@ -63,33 +66,18 @@ metadata:
 spec:
   description: |-
     A summary of the resource
-
+
```
sthaha, line 69: nit: whitespace 🙏

sthaha (commented):
Looks good to me but I will let @vdemeester and @chmouel review and merge.

catlin/go.mod
```diff
@@ -3,16 +3,13 @@ module github.com/tektoncd/plumbing/catlin
 go 1.14
 
 require (
+	github.com/google/go-containerregistry v0.1.1
 	github.com/spf13/cobra v1.0.0
 	github.com/tektoncd/pipeline v0.15.2
 	go.uber.org/zap v1.15.0
 	gotest.tools/v3 v3.0.2
-
```
sthaha, line 10: could you please revert back unrelated change?

catlin/pkg/validator/task_validator_test.go
```diff
@@ -0,0 +1,190 @@
+package validator
```
sthaha, line 1: license 🙏

catlin/go.mod
SM43: `go mod tidy` removed them I guess :thinking:

catlin/pkg/cmd/validate/validate.go
```diff
@@ -84,10 +87,18 @@ func validate(cli app.CLI, path string) error {
...
 	out := cli.Stream().Out
 	for _, v := range result.Lints {
-		if v.Kind == validator.Error {
+		switch v.Kind {
+		case validator.Error:
 			fmt.Fprintf(out, "ERROR: %s\n", v.Message)
-		} else {
+		case validator.Warning:
+			fmt.Fprintf(out, "WARN : %s\n", v.Message)
+		case validator.Recommendation:
```
sthaha, line 95: unrelated to this patch: perhaps we should rename `Recommendation` to `Hint` :D

SM43: yeah :smiley_cat: will address in a separate patch

vdemeester (approved):
/meow

Issue #565, opened 2020-09-10
Catlin: add catlin validation to tektoncd/catalog PR


sthaha:
Some of the tasks in tektoncd/catalog doesn't comply with the Re-Org TEP. Let's add catlin to the Catalog CI so that all new submissions pass the `catin` validation. 

/kind feature

PR #584, opened 2020-09-22
Catlin: Adds Dockerfile

SM43:
Part of #565 

Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

# Submitter Checklist

/kind misc

### DIFF
--- a/catlin/Dockerfile
+++ b/catlin/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:1.14 as BUILD
+
+COPY . /build
+WORKDIR /build
+RUN GOOS=linux GARCH=amd64 CGO_ENABLED=0 \
+    go build -o catlin ./cmd/catlin
+
+FROM scratch
+
+WORKDIR /app
+COPY --from=BUILD /build/catlin /app/catlin
+
+ENTRYPOINT ["/app/catlin"]
+


SM43:
@vdemeester @afrittoli @sthaha  Could you please review this? :smiley_cat:

sthaha (commented):
/lgtm

Will let @vdemeester approve.

catlin/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.14 as BUILD
+
+COPY . /build
+WORKDIR /build
+RUN GOOS=linux GARCH=amd64 CGO_ENABLED=0 \
+    go build -mod=vendor -o catlin ./cmd/catlin
+
+FROM scratch
```
vdemeester, line 8: nit: any reason to use `scratch` ? long-term we may need to want to debug a container running this image, and/or might need to stuff like certificates, …
Any way, no need to change this now

vdemeester (commented):
@SM43 any reason to vendor the dependencies ? we do have multiple go modules in the repository, and so far we didn't use `vendor` for them :sweat_smile: 
If we want to go that way (vendoring), we would need a script to make sure the `vendor` folder is up-to-date. I would rather not vendor anything, at least not while `catlin` is in this repository.

catlin/Dockerfile
sthaha: > @SM43 any reason to vendor the dependencies ? we do have multiple go modules in the repository, and so far we didn't use `vendor` for them 😅
> If we want to go that way (vendoring), we would need a script to make sure the `vendor` folder is up-to-date. I would rather not vendor anything, at least not while `catlin` is in this repository.

Lets go without vendor then :D ... I am not a big fan of vendor either but the image build is super fast 😝

sthaha (commented):
/lgtm
/approve
/meow

Issue #592, opened 2020-09-24
Multi-architecture build of e2e and plumbing images


vdemeester:
In order to [support multiple architecture](https://github.com/tektoncd/pipeline/issues/856), we need to provide image in multiple
architecture for the projects's end-to-end tests. This also applies to some images built
in plumbing as we may use them on the target clusters (s390x, arm64, …) and/or during
end-to-end tests.

In an ideal world, we would use tekton to build those (aka dogfooding, …). My initial
search around this mainly ended up around `qemu`, `qemu-user-static` and `docker` (+buildkit).

See

-   [crazy-max/ghaction-docker-buildx: GitHub Action to set up Docker Buildx](https://github.com/crazy-max/ghaction-docker-buildx)
-   [hello-multiarch/image.yml at master · piwi3910/hello-multiarch](https://github.com/piwi3910/hello-multiarch/blob/master/.github/workflows/image.yml)

So the questions is how should we proceed:

-   Have a quick way of building multi-arch images using github action (as above)
-   Have a "middle" solution where we use dogfooding with some DinD, privileged containers
    and buildx
-   Find some way to replicate the github action flow in tekton (ideally without any need to
    privileged containers, but I have high doubt that it's possible)

/cc @afrittoli 
/kind feature

barthy1:
Can first versions of the mentioned images be built using the fastest way? I guess it would be github actions. Just to unblock "other arch support" activities..
And some longer but more conceptual approach can be used after.

vdemeester:
@barthy1 yeah, @afrittoli said the same during our productivity working group :angel:

barthy1:
@vdemeester I'd like to suggest the first candidate for multi-arch build here is `gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest`, as it's used in [.ko.yaml ](https://github.com/tektoncd/pipeline/blob/master/.ko.yaml) for pipeline, and required to be mluti-arch to use `--platform=all` option. At this moment it is built by [gcr.io/kaniko-project/executor:v0.17.1](https://github.com/tektoncd/pipeline/blob/master/tekton/publish.yaml#L53), which doesn't support [multi-arch manifest](https://github.com/GoogleContainerTools/kaniko/issues/1102).
Do you think it makes sense to have the workflow to run Github actions build in [plumbing](https://github.com/tektoncd/plumbing) or better in [pipeline](https://github.com/tektoncd/pipeline)?

barthy1:
[...]

PR #603, opened 2020-10-05
Add github action workflow to build pipeline base images

barthy1:
# Changes

Currently pipeline [base image](https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L5) is built using [gcr.io/kaniko-project/executor:v0.17.1](https://github.com/tektoncd/pipeline/blob/master/tekton/publish.yaml#L53) which doesn't support multi-arch builds, so it is a stopper to get multi-arch pipeline releases.

This Github actions workflow can be used to build pipeline base image in multi-arch form with `docker buildx`.
Architectures for manifest are amd64, arm64, s390x, ppc64le.
The build runs on amd64 architecture, executing builds for other architectures using emulation.

The setup requires to have 2 secret parameters("GKE_SA_KEY" and "GKE_PROJECT")
to push images to Google Cloud registry.
There are also 3 input parameters to construct image name. The default values will create the `gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest` image.
Workflow can be start manually at github.com or via Github API.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/.github/workflows/pipeline_base_image_build.yml
+++ b/.github/workflows/pipeline_base_image_build.yml
@@ -1,0 +1,63 @@
+name: build pipeline base image
+
+on:
+  workflow_dispatch:
+    inputs:
+      pathToProject:
+        description: 'path to project'
+        required: true
+        default: 'github.com/tektoncd/pipeline'
+      imageRegistry:
+        description: 'image registry'
+        required: true
+        default: 'gcr.io/tekton-nightly'
+      url:
+        description: 'url'
+        required: true
+        default: 'build-base:latest'
+
+jobs:
+  buildx:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: github_actions_base_pipeline
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+          install: true
+
+      - name: Set up gcloud
+        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: '290.0.1'
+          service_account_key: ${{ secrets.GKE_SA_KEY }}
+          project_id: ${{ secrets.GKE_PROJECT }}
+
+      - name: Configure docker to use gcloud
+        run: |-
+          gcloud --quiet auth configure-docker
+
+      - name: Checkout pipeline repo
+        uses: actions/checkout@v2
+        with:
+          repository: tektoncd/pipeline
+          path: pipeline
+
+      - name: Run Buildx
+        run: |
+          docker  build \
+            --platform linux/amd64,linux/arm64,linux/ppc64le,linux/s390x \
+            --tag ${{ github.event.inputs.imageRegistry }}/${{ github.event.inputs.pathToProject }}/${{ github.event.inputs.url }} \
+            --push \
+            ./pipeline/images


bobcatfish:
I think @vdemeester or @afrittoli were telling me about this recently but I'm wondering if using GitHub actions is what is solving our problem here or if it is the fact that the GitHub action(s) are using something other than kaniko to build with? There is no reason that we _have_ to keep using kaniko - if using a different image / build method solves this problem we could use it.

(Basically I'm trying to make sure that there isn't some underlying issue with Tekton itself that makes Tekton not work for this scenario, because if that _is_ the case I think it's something we'd want to fix)

vdemeester:
> I think @vdemeester or @afrittoli were telling me about this recently but I'm wondering if using GitHub actions is what is solving our problem here or if it is the fact that the GitHub action(s) are using something other than kaniko to build with? There is no reason that we _have_ to keep using kaniko - if using a different image / build method solves this problem we could use it.
> 
> (Basically I'm trying to make sure that there isn't some underlying issue with Tekton itself that makes Tekton not work for this scenario, because if that _is_ the case I think it's something we'd want to fix)

This comes from https://github.com/tektoncd/plumbing/issues/592. We discussed this in the productivity call and we kinda decided to go the quick way to unlock this while we work on making progress (:crossed_fingers:) on having all this in Tekton. The current "problem" with tekton is not from tekton itself but more on "building multi-arch images in kubernetes", it doesn't have anything to do with the architecture of tekton. As far as I know, there is not way to build multi-arch images using `kaniko` in a generic way. The current "way" to do it with `docker` (and `buildkit`) is to use qemu and some qemu-user-static _hack_ that comes with a bunch of problems to run it in a kubernetes cluster (requires `privileged`, some assumption on the kernel, …). GitHub Actions do not have this problem because they are VMs and basically, you can do whatever in (because you can be root in there).
We could run VMs in kubernetes and do something similar to GitHub actions, but it might take some time — Can this be done in GKE, ….

We *can* wait to run all this in tekton, or we can do the same as we do with prow, use what works while we wait and work on a solution completely tekton based :angel:

.github/workflows/pipeline_base_image_build.yml
```diff
@@ -0,0 +1,63 @@
...
+          repository: tektoncd/pipeline
+          path: pipeline
+
+      - name: Run Buildx
+        run: |
+          docker  build \
+            --platform linux/amd64,linux/arm64,linux/ppc64le,linux/s390x \
+            --tag ${{ github.event.inputs.imageRegistry }}/${{ github.event.inputs.pathToProject }}/${{ github.event.inputs.url }} \
+            --push \
+            ./pipeline/images
```
vdemeester, line 63: Shouldn't this be different ? :thinking:

barthy1: you mean platforms? which one for you suggest?

vdemeester: nah I meant the `./pipeline/image` path. where is it in the `tektoncd/plumbing` repository ? shouldn't this be pointing at each images with a Dockerfile ?

barthy1: so far I've created this PR to build 1 image -`gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest `- https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L5
And I thought its Dockerfile is located at https://github.com/tektoncd/pipeline/tree/master/images, so `pipeline/images` for pipeline repo

barthy1: if it works smoothly we can build other images from plumbing repo itself

vdemeester: ohhhh It's targeting `tektoncd/pipeline`, I see :sweat_smile:

bobcatfish:
Thanks for the explanation @vdemeester , I agree with moving ahead to unblock this (and think it's important we document why we are making these kinds of decisions).

ImJasonH:
I'd like to avoid taking a dependency on GitHub Actions if we can avoid it, since our CI/CD setup is already sort of complicated, involving both Prow and dogfooded Tekton workflows.

If the issue at hand is that kaniko is incapable of building multi-arch images, I think there are two ways we could get around it while staying within Tekton:

1. Make kaniko capable of building multi-arch images. If not in the general case, at least in the specific case we'd need to unblock ourselves -- it looks like we just need to install some packages, we don't even need to `RUN` anything besides that. @mattmoor has done some exploration here, and I think it might be closer than we thinkn
1. Use `docker build` with a dind sidecar instead of kaniko, at least for now. I suspect this is along the lines of what GitHub Actions is doing under the hood (using `docker build`, unclear about whether dind is involved).

If those options have already been discussed and rejected it would be good to understand why.

mattmoor:
@ImJasonH I posted the changes I used to build a multi-arch kaniko image here: https://github.com/GoogleContainerTools/kaniko/pull/1452

The 0.18 release of mink I cut last night has builds of everything (incl. tekton and kaniko) with arm64: https://github.com/mattmoor/mink/releases/tag/v0.18.0

These aren't producing multi-arch images themselves, but since kaniko can now run on arm64 it can produce arm64 images, and then these can be stitched together with `docker manifest create`.  Perhaps you could run a multi-arch cluster with amd64 and arm64 nodepools and use a Pipeline to orchestrate Tasks for each architecture w/ nodeSelectors, and then a final task that stitches them all together.

It'd be a really powerful pattern to demonstrate with Tekton (w/ sample, blog), but I have not tried it.

barthy1:
@mattmoor agree that having just heterogeneous cluster with workers on different architectures is very interesting use case. And I am really looking forward to see that (or even participate with workers on ppc64le or s390x). However GKE where the Tekton dogfooding cluster is running has only amd64 clusters (please, correct me if I am wrong).

At this moment the question is how to build `build-base` image for many architectures to be able to release Tekton for these architectures :) Without Tekton build for specific arch you cannot run it on worker with specific arch.

I'd like also to point out that `bazel` which was used to build `kaniko` [here](https://github.com/GoogleContainerTools/kaniko/pull/1452/files) is released at this moment only for `arm64` and `amd64`. So `s390x`, `ppc64le` and other possible architectures will be excluded in this picture, because of building tool dependency.

mattmoor:
> I'd like also to point out that bazel which was used to build kaniko here is released at this moment only for arm64 and amd64. So s390x, ppc64le and other possible architectures will be excluded in this picture, because of building tool dependency.

Test dependency yet, but since Bazel can cross-compile images, getting the builds effectively devolves to adding those architectures here: https://github.com/GoogleContainerTools/kaniko/pull/1452/files#diff-653e8b99fa9eba9096b9386de586f579R19-R22

mattmoor:
> However GKE where the Tekton dogfooding cluster is running has only amd64 clusters (please, correct me if I am wrong).

Not wrong, I just wasn't limiting the infra to GKE.  I have no clue how Tekton operates (or funds) its infra, so 🤷

barthy1:
> Test dependency yes, but since Bazel can cross-compile images, getting the builds effectively devolves to adding those 
 architectures here: https://github.com/GoogleContainerTools/kaniko/pull/1452/files#diff-653e8b99fa9eba9096b9386de586f579R19-R22

Thanks for pointing me to that, as `kaniko/executor` is in our plans for `s390x`, because it is used for many Tekton tests anyway.
So if we can just cross compile it, it sounds really promising 👍

mattmoor:
@barthy1 Yeah, no problem.  I think the infra question remains because we need a place to *run* kaniko natively for that to work, but I'd like to push kaniko upstream to publish a wider variety of architectures once we have some miles through the amd64/arm64 version.  After all, there's effectively nothing architecture specific in that image other than the binary, and all we're doing is cross-compiling which is a pretty well exercised path, I'd think.

barthy1:
> I think the infra question remains because we need a place to run kaniko natively for that to work

Well :) we can use Travis with its native hardware, but it's still extension and using another CI tool :(

Github Actions was suggested just as faster and working way to speed up multi-arch builds at first step with plan in the future to switch to Tekton Tasks anyway. However I think this discussion is really great to collect some other ideas and have opportunity to verify them..

mattmoor:
As a fairly radical alternative, you could try to eliminate the Dockerfile dependency entirely by replacing the tooling dependencies with libraries.

e.g. where `git` is used today start to adopt https://godoc.org/gopkg.in/src-d/go-git.v4 etc

vdemeester:
> I'd like to avoid taking a dependency on GitHub Actions if we can avoid it, since our CI/CD setup is already sort of complicated, involving both Prow and dogfooded Tekton workflows.
>
> […]
> 
> If those options have already been discussed and rejected it would be good to understand why.

In general, we all agree it would be better to not depend on any other CI *and* we should aim to build anything in tekton. That said, building multiarch images can be tricky to achieve, in kubernetes at least (aka in containers).

> If the issue at hand is that kaniko is incapable of building multi-arch images, I think there are two ways we could get around it while staying within Tekton:
>     1. Make kaniko capable of building multi-arch images. If not in the general case, at least in the specific case we'd need to unblock ourselves -- it looks like we just need to install some packages, we don't even need to `RUN` anything besides that. @mattmoor has done some exploration here, and I think it might be closer than we thinkn
>     2. Use `docker build` with a dind sidecar instead of kaniko, at least for now. I suspect this is along the lines of what GitHub Actions is doing under the hood (using `docker build`, unclear about whether dind is involved).

This is, indeed, one issue, but let's give a bit more context. This is mainly about #592, and thus about **building `Dockerfile` for multiarch**. We *can't* make any assumption on the content of the `Dockerfile` (`test-runner` does quite some `RUN` and things for example). Indeed, (1) is possible, *but* it needs time and this would make `kaniko` run on multiple architecture, which means we need *hardware* (or a VM) to run it for the given architecure, which gets us to (2). To build multiarch images using `docker` (2), it is more involved than a normal `docker build` (there is no magic). We basically need to do what [this github action does](https://github.com/docker/setup-qemu-action), aka, run [something](https://github.com/multiarch/qemu-user-static) that will setup `binfmt`, … to execute the correct qemu static binary to run with the given architecture. In a cluster, this is very involved, it needs privileged access and may setup the node kernel in some weird way. I did not try to do this with `DinD` but still, if it works, it will still need privileged.

> As a fairly radical alternative, you could try to eliminate the Dockerfile dependency entirely by replacing the tooling dependencies with libraries.

This works for `tektoncd/pipeline` and go projects, it definitely doesn't for [`./tekton/images`](https://github.com/tektoncd/plumbing/tree/master/tekton/images) in here.

In the light of the comments, I feel @chmouel's [comment](https://github.com/tektoncd/plumbing/issues/592#issuecomment-705567257) is a good middle ground. We can create tasks to create a gcloud compute node, setup docker on it, build the images using docker and get the result.

barthy1:
Close this PR in favour of https://github.com/tektoncd/pipeline/pull/3402

PR #613, opened 2020-10-15
Remove myself from the build cop rotation 🧘❤️

bobcatfish:
# Changes

Lately I've been looking for ways to lighten my load. In stepping back
and looking for things I could stop doing, it feels reasonable to me
for me to stop build copping, at least for a while, so I'd like to
remove myself from the rotation.

I've made the change after next week so I can do one more rotation and
because this changes the entire schedule after that point so I want to
minimize the disruptions to everyone else.

Thanks! ❤️

This this changes the whole rotation after that point, heads up @afrittoli @wlynch @dibyom @nikhil-thomas @savitaashture @sbwsg @jerop @vdemeester 

# Submitter Checklist

review them:

for more details._


bots/buildbot/rotation.csv
```diff
@@ -108,44 +11,45 @@ Date,User
 2020-10-21,wlynch
 2020-10-22,christiewilson
 2020-10-23,dibyo
+Date,User
```
sbwsg, line 14: this is interesting - the Date,User heading appears twice in the file now.  must be a bug in the rotation generator :/

sbwsg (approved):

bots/buildbot/rotation.csv
bobcatfish: whoops, thanks for catching that!

vdemeester (commented):
/lgtm
/meow

PR #628, opened 2020-11-05
Add nightly release for tekton website docs for master version

popcor255:
# Changes

The Tekton Website uses github actions to sync the master version of the docs. We should have this sync on the tekton plumbing repo. The reason why this is the best approach is because it consolidates our infra and gives us an opportunity to dogfood. We currently sync on github actions as of https://github.com/tektoncd/website/commit/428238c3e9b42834ba663f76179e1cffc7b3243a. This website will only sync at 3:15AM UTC. This means that master version of the docs will not get updated until that time. A secret with the netlify build hook needs to be added to the cluster for the cronjob to work properly.

# Submitter Checklist

review them:

for more details._

/area testing
/kind misc

### DIFF
--- a/tekton/cronjobs/dogfooding/tekton/kustomization.yaml
+++ b/tekton/cronjobs/dogfooding/tekton/kustomization.yaml
@@ -2,3 +2,4 @@ resources:
 - dashboard-to-robocat-nightly
 - pipeline-to-robocat-nightly
 - triggers-to-robotcat-nightly
+- website-to-netlify-nightly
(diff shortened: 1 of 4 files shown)


afrittoli (commented):
Thanks for this!
This does not use a Tekton task which means that the execution log is not publicly available (via the dashboard). I guess that's ok for now, we can still run this into a trigger to a task as a second step.

The only request I have before it's merged is if you could add a comment on https://github.com/tektoncd/plumbing/blob/master/docs/dogfooding.md#secrets so that we keep track of secrets required to provision the cluster.

popcor255:
/hold
I think it might be better to sync the site generic cd-pipeline rather than having a job.

PR #627, opened 2020-11-03
Add the buildx cronjob to kustomize so it gets deployed

barthy1:
# Changes

Add `buildx-gcloud-nightly` to kustomization list.
Also update documentation how to add daily builds for new images.

/kind misc

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #629, opened 2020-11-06
Add build of dogfooding images in multi-arch form

barthy1:
# Changes

Extend current cronjobs/triggers setup to build dogfooding images in multi-arch form when it's necessary:
- create 2 TriggerTemplates and TriggerBindings for single-arch and multi-arch cases
- extend eventlistener to decide which build to do (based on platforms parameter value)
- add PLATFORMS parameter to cronjobs to be able to specify for which platforms multi-arch image should be
- extend several image cronjobs to use multi-arch build (`tkn`, `alpine-git-nonroot`, `skopeo`)
- for multi-arch build `buildx` in main container and `dind` sidecar are used
- builds will be done on amd64 dogfooding cluster, not other native hardware is required.

Not every image can be built with this approach, so I suggest to use it only for images, which are required for the tests.
The build cronjobs were updated to those exact images.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cronjobs/dogfooding/images/skopeo-nightly/cronjob.yaml
+++ b/tekton/cronjobs/dogfooding/images/skopeo-nightly/cronjob.yaml
@@ -22,3 +22,5 @@ spec:
               value: gcr.io/tekton-releases/dogfooding/skopeo:latest
             - name: CONTEXT_PATH
               value: tekton/images/skopeo
+            - name: PLATFORMS
+              value: "linux/amd64,linux/s390x,linux/ppc64le,linux/arm64"
--- a/tekton/resources/images/bindings.yaml
+++ b/tekton/resources/images/bindings.yaml
@@ -1,0 +1,31 @@
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: trigger-to-build-and-push-image
+spec:
+  params:
+  - name: buildUUID
+    value: $(body.buildUUID)
+  - name: gitRepository
+    value: $(body.gitRepository)
+  - name: gitRevision
+    value: $(body.gitRevision)
+  - name: contextPath
+    value: $(body.contextPath)
+  - name: registry
+    value: $(body.registry)
+  - name: namespace
+    value: $(body.namespace)
+  - name: imageName
+    value: $(body.imageName)
+  - name: imageTag
+    value: $(body.imageTag)
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: trigger-to-build-and-push-image-platform
+spec:
+  params:
+  - name: platforms
+    value: $(body.platforms)
--- a/tekton/resources/images/eventlistener.yaml
+++ b/tekton/resources/images/eventlistener.yaml
@@ -1,0 +1,28 @@
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: image-builder
+spec:
+  serviceAccountName: release-right-meow
+  triggers:
+    - name: single-arch-build-trigger
+      interceptors:
+        - cel:
+            filter: >-
+              !('platforms' in body) ||
+              size(body['platforms']) == 0
+      bindings:
+        - ref: trigger-to-build-and-push-image
+      template:
+        name: build-and-push-image
+    - name: multi-arch-build-trigger
+      interceptors:
+        - cel:
+            filter: >-
+              'platforms' in body &&
+              size(body['platforms']) != 0
+      bindings:
+        - ref: trigger-to-build-and-push-image
+        - ref: trigger-to-build-and-push-image-platform
+      template:
+        name: build-and-push-image-multi-arch
--- a/tekton/cronjobs/bases/image-build/trigger-image-build.yaml
+++ b/tekton/cronjobs/bases/image-build/trigger-image-build.yaml
@@ -64,7 +64,8 @@ spec:
                   "registry": "$REGISTRY",
                   "namespace": "$NAMESPACE",
                   "imageName": "$IMAGE",
-                  "imageTag": "$TAG"
+                  "imageTag": "$TAG",
+                  "platforms": "$PLATFORMS"
                 }
                 EOF
                 curl -d @/workspace/post-body.json $SINK_URL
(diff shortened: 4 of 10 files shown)


barthy1:
cc @afrittoli

tekton/cronjobs/dogfooding/images/skopeo-nightly/cronjob.yaml
```diff
@@ -22,3 +22,5 @@ spec:
               value: gcr.io/tekton-releases/dogfooding/skopeo:latest
             - name: CONTEXT_PATH
               value: tekton/images/skopeo
+            - name: PLATFORMS
+              value: "linux/amd64,linux/s390x,linux/ppc64le,linux/arm64"
```
savitaashture, line 26: I see platform `linux/arm64` in few yamls any reason for that ?

vdemeester (approved):
/cc @afrittoli

tekton/cronjobs/dogfooding/images/skopeo-nightly/cronjob.yaml
barthy1: At this moment pipeline nightly builds are done for amd64, s390x, ppc64le and arm64. So I thought it would be nice to have also arm64 dogfooding images where it's possible. For some images it's not possible, because proposed multi-arch build requires corresponding availability of base images for same architectures, and for arm64 some base images are not available

tekton/resources/images/overlays/multi-arch/image-build-trigger-template.yaml
```diff
@@ -0,0 +1,70 @@
+- op: add
+  path: "/spec/resourcetemplates/0/spec/taskSpec/steps"
+  value:
+  - image: gcr.io/tekton-releases/dogfooding/buildx-gcloud:mario
```
afrittoli, line 4: The `mario` tag is just for testing, we should not use it in here. Since we do nightly builds, you could use `latest` or a specific nightly tag/sha

tekton/resources/images/overlays/multi-arch/image-build-trigger-template.yaml
```diff
@@ -0,0 +1,70 @@
+- op: add
+  path: "/spec/resourcetemplates/0/spec/taskSpec/steps"
```
afrittoli, line 2: NIT: I'm not a huge fan of the idea of injecting steps into a common task here, as it may make it more difficult to make changes to one of the tasks later on (like switching to catalog tasks, or building tasks to a bundle), but that's also something we can also change later when the time comes.

tekton/resources/images/overlays/multi-arch/configuration.yaml
```diff
@@ -0,0 +1,9 @@
+nameReference:
+- kind: TriggerBinding
```
afrittoli, line 2: NIT: There's no need to use kustomize on bindings, you can specify multiple bindings in a trigger, so for me the way to go would be to have the common base binding and another platform binding which is only used on the multi-arch trigger.

tekton/resources/images/overlays/multi-arch/configuration.yaml
```diff
@@ -0,0 +1,9 @@
...
+  fieldSpecs:
+  - path: spec/triggers/bindings/ref
+    kind: EventListener
+- kind: TriggerTemplate
+  fieldSpecs:
+  - path: spec/triggers/template/name
+    kind: EventListener
```
afrittoli, line 9: NIT: I've been trying to keep the number of event listeners to a minimum, I feel that having too many services running makes it more difficult to manage. The main reason I kept this event listener separated from the others is that this one uses a service account that can push to the container registry. 
All other CD services are in the same event listener, each matches its own trigger, and the correct trigger is selected using a CEL filter on a dedicated field in the payload sent by the cronjob.

afrittoli (commented):
Thanks for this!
I have a few comments, but they're not really blocking, more like my personal opinions - it would be nice to change the image not to use the `mario` tag though, since that is only for testing and it may be overwritten.

tekton/resources/images/overlays/multi-arch/image-build-trigger-template.yaml
barthy1: we don't have nightly build so far, as https://github.com/tektoncd/plumbing/pull/627 is not merged :(
so `gcr.io/tekton-releases/dogfooding/buildx-gcloud:mario` is only one publicly available version

tekton/resources/images/overlays/multi-arch/image-build-trigger-template.yaml
barthy1: I was thinking if we can have 2 separate sets for multi-arch and single-arch cases. But decided that having just separate code, where many parts are the same would lead even to more problems, when something needs to be updated for both cases. 
However if based on your experience it's better to do, I'd be happy to follow this path

tekton/resources/images/overlays/multi-arch/configuration.yaml
barthy1: So, if I understand everything correct, one EventListener for multi-arch and single-arch cases, different TriggerTemplates and TriggerBindings, use CEL filter to decide which trigger.. to use :) Will work on it

tekton/resources/images/bindings.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - name: contextPath
+    value: $(body.contextPath)
+  - name: registry
+    value: $(body.registry)
+  - name: namespace
+    value: $(body.namespace)
+  - name: imageName
+    value: $(body.imageName)
+  - name: imageTag
+    value: $(body.imageTag)
```
afrittoli, line 45: Could you keep only `platforms` in this binding, and add both bindings in the event listener for the multi-arch case?

tekton/resources/images/eventlistener.yaml
```diff
@@ -0,0 +1,27 @@
...
+      template:
+        name: build-and-push-image
+    - name: multi-arch-build-trigger
+      interceptors:
+        - cel:
+            filter: >-
+              'platforms' in body &&
+              size(body['platforms']) != 0
+      bindings:
+        - ref: trigger-to-build-and-push-image-multi-arch
```
afrittoli, line 25: It'd be nice to have a list of two here:
```
- trigger-to-build-and-push-image
- trigger-to-build-and-push-image-platform
```

afrittoli (commented):
Thanks for all the updates!
One last thing and then it's ready to go.

afrittoli (commented):
/approve
/lgtm

PR #637, opened 2020-11-13
Add bef args as valid to deploy-release script

sbwsg:
# Changes

We were calling `getopts` with only `:p:v:` which means
only the -p and -v flags were considered valid.

This commit expands the set of valid flags to include b,e,f.

# Submitter Checklist

review them:


afrittoli (commented):
Nice catch, thank you!
/approve

vdemeester (approved):
/lgtm

PR #639, opened 2020-11-16
Support e2e tests with kind

### LINKED ISSUE #631: Support e2e testing with kind
### Feature request

Currently e2e tests [create a regional GKE cluster using kubetest](https://github.com/tektoncd/plumbing/blob/b086b5c7ad03e01520aa166b3f1a97fd643ceb54/scripts/e2e-tests.sh#L251-L256) -- this takes a few minutes to set up, and can cost money if resources leak. Requiring a GCP account and billing might present a barrier to some contributors. We also lose some control over the version of Kubernetes installed on the GKE cluster, since GKE doesn't support every old version of Kubernetes, or immediately support new versions as they're released. We're also limited by GKE's chosen cluster configuration and feature flags.

We should support an e2e testing mode using [`kind`](https://kind.sigs.k8s.io/), which simulates a K8s cluster using local Docker containers. `kind cluster create` typically finishes in a few seconds, and [lets you configure a K8s version](https://kind.sigs.k8s.io/docs/user/quick-start/#setting-kubernetes-version), [configure feature flags](https://kind.sigs.k8s.io/docs/user/quick-start/#enable-feature-gates-in-your-cluster), and more.

Though they look like multi-node clusters, all resources are consumed from the machine running the cluster, so this might limit the size of the cluster we can create, especially for local testing. We shouldn't remove the GKE option, and should continue to run all CI tests against GKE, in addition to kind.

### Use case

Since clusters are easier to create and destroy, we could write tests against a small kind cluster to test resource-exhaustion scenarios, tests that target specific K8s versions or feature flag configurations, install mutating webhooks to inject sidecars, etc.

### Prior Art

Knative runs tests in kind, see https://github.com/knative/hack/blob/master/test/e2e-kind.sh and Prow changes [here](https://github.com/knative/test-infra/pull/2427).

@mattmoor has done a lot of kind testing for mink, and might be able to provide guidance if we get stuck.

@afrittoli has contributed [a script in this repo](https://github.com/tektoncd/plumbing/blob/master/hack/tekton_in_kind.sh) to create a kind cluster with Tekton installed.

yaoxiaoqi:
# Changes

This PR closes tektoncd#631.

This PR supports running e2e tests with kind. The kind cluster runs locally so that a GCP account is not required
to perform e2e tests. When the script is set to kind mode, KO_DOCKER_REPO will be `kind.local` in order to install
tekton pipelines controller onto the kind cluster.

# Submitter Checklist

review them:

- [] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -152,7 +152,8 @@ function delete_leaked_network_resources() {
   fi
 }
 
-# Create a test cluster with kubetest and call the current script again.
+# Create a test cluster and call the current script again.
+# Parameters: $1 - cluster provider name, e.g. gke
 function create_test_cluster() {
   # Fail fast during setup.
   set -o errexit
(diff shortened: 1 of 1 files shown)


afrittoli:
Thanks for this PR.

My intention for kind testing was to set this so to that it would run in Tekton, mainly because I always want to do more dogfooding. That said I'm not opposed on adding support via prow/kubetest2 as that may give us a quicker route into testing against multiple version of k8s.

Do you know the details of how kubetest2 setup kind on k8s?

yaoxiaoqi:
@afrittoli I have read the reference link you commented in the issue and guess you might want to run kind in a k8s cluster, like creating a k8s cluster inside a k8s cluster if I understand correctly. Unfortunately, I don't think kubetest2 can set up kind inside a kubernetes cluster. This PR mainly supports running e2e tests locally by using kind so that people can run integration tests without a GCP account. But I think your suggestion is useful and can be implemented in the future.

ImJasonH:
> Thanks for this PR.
> 
> My intention for kind testing was to set this so to that it would run in Tekton, mainly because I always want to do more dogfooding. That said I'm not opposed on adding support via prow/kubetest2 as that may give us a quicker route into testing against multiple version of k8s.

I think we should consider _both_, as separate streams of work. I think there's definitely value in providing a `kind-test` Task, or `create-kind-cluster` and `delete-kind-cluster` Tasks or similar, and if those existed we could maybe even dogfood them instead of using Prow.

But I think we also want to be able to use kind without having to wait for those Tasks to exist, and I think there's value in having kind-in-Prow work.

> 
> We would need a test image with `kubetest2` installed though.
> Do you know the details of how kubetest2 sets up kind on k8s?

We could also [install kubetest2 on-demand](https://github.com/kubernetes-sigs/kubetest2#installation) as part of e2e-common.sh as a short-term solution, then move it into the test image once we're sure the tests are reliable.

yaoxiaoqi:
Hi @ImJasonH , could you help me review this PR?

PR #647, opened 2020-11-24
scripts: Make sure `cat` params are quoted.

wlynch:
# Changes

If `cat` is called with no params, it assumes that it needs to read from
stdin, and will block for user input. This is problematic if
${CHANGED_FILES} is unset (i.e. when not running in a PR), since it
looks like the script just hangs.

Long term we should probably look into modifying the ${CHANGED_FILES}
behavior to fallback to a local diff, but for now make sure params are
quoted so that the script fails fast.

# Submitter Checklist

review them:

for more details._


vdemeester (approved):
/lgtm

PR #692, opened 2021-01-22
Add Makefile for Catlin

chmouel:
# Changes

Add a Makefile for catlin with a few common targets.
Add the vendor directory too.

# Submitter Checklist

review them:

for more details._


vdemeester:
I wonder if we want to add `vendor` though. So far we didn't rely on `vendor` for any go tooling in that repository to not have a huge content (like multiple vendor directory for each go tool). @chmouel Could we split the `Makefile` from the vendor part to get one in quickly and discuss the other ?

chmouel:
either way to not adding it i am fine... i just followed what we had in other tekton repository.

bobcatfish:
@chmouel is this the kind of thing we'd want to have in our Pipelines and Tasks? i'm wondering why we'd want to use a makefile for this vs having using a go build Task

chmouel:
@bobcatfish it's only there to be able run the test or linter easily when developing on the local developer machine, I don't think we have a solution for that

bobcatfish:
ah okay, interesting, thanks @chmouel

i hope one day we'll make it so we don't have to have a different set of scripts and automation for local development (https://github.com/tektoncd/community/issues/145 🤞 )

chmouel:
I think the yamllint job has some issues i have logged an issue here : 

https://github.com/tektoncd/plumbing/issues/698

afrittoli (commented):
/approve

vdemeester (approved):
/lgtm

PR #696, opened 2021-01-25
Hot reload the build captain csv 🔥

bobcatfish:
# Changes

While updating build cop -> build captain, I noticed (which I think
other folks have been saying all along!) that the list of ppl in the csv
is loaded once on startup, meaning if there are any updates, you have to
restart to bot. With this change, the list will be loaded every time it
is accessed.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/bots/buildcaptain/cmd/buildcaptain/main.go
+++ b/bots/buildcaptain/cmd/buildcaptain/main.go
@@ -13,7 +13,6 @@ import (
 const rotationURL = "https://raw.githubusercontent.com/tektoncd/plumbing/master/bots/buildcaptain/rotation.csv"
 
 var (
-	currentCaptain string
 	botID          string
 	token          string
 	channelID      string
@@ -51,12 +50,9 @@ func main() {
 		captainsID[user.Name] = user.ID
 	}
 
-	r := NewRotation(FromURL(rotationURL))
-	currentCaptain = captainsID[r.GetBuildCaptain(time.Now())]
-
 	rtm := api.NewRTM()
 	go rtm.ManageConnection()
-	go dailyPing(rtm, captainsID, r)
+	go dailyPing(rtm, captainsID)
 
 	for msg := range rtm.IncomingEvents {
 		switch ev := msg.Data.(type) {
@@ -67,7 +63,7 @@ func main() {
 			fmt.Println("Connection counter:", ev.ConnectionCount)
 		case *slack.MessageEvent:
 			if isDirectMessage(ev.Channel) || ev.Channel == channelID {
-				handleMessage(rtm, ev.Text, ev.Channel, isDirectMessage(ev.Channel))
+				handleMessage(rtm, ev.Text, ev.Channel, isDirectMessage(ev.Channel), captainsID)
 			}
 		case *slack.PresenceChangeEvent:
 			fmt.Printf("Presence Change: %v\n", ev)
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/meow
/cc @afrittoli @sbwsg

bots/buildcaptain/cmd/buildcaptain/main.go
```diff
@@ -85,10 +81,10 @@ func main() {
 	}
 }
 
-func handleMessage(rtm *slack.RTM, message, channel string, direct bool) {
+func handleMessage(rtm *slack.RTM, message, channel string, direct bool, captainsID map[string]string) {
 	switch {
 	case statusMessage(message, botID, direct):
-		rtm.SendMessage(rtm.NewOutgoingMessage(fmt.Sprintf("<@%s> is the Build Captain :female-pilot:\nBuild Captain log is here: https://docs.google.com/document/d/1kUzH8SV4coOabXLntPA1QI01lbad3Y1wP5BVyh4qzmk", currentCaptain), channel))
+		rtm.SendMessage(rtm.NewOutgoingMessage(fmt.Sprintf("<@%s> is the Build Captain :female-pilot:\nBuild Captain log is here: https://docs.google.com/document/d/1kUzH8SV4coOabXLntPA1QI01lbad3Y1wP5BVyh4qzmk", getCurrentCaptain(captainsID)), channel))
```
sbwsg, line 87: Am I understanding right that every time I request "status" from the slack bot we're going to make a call to github?

That seems like a vector for a bad actor to take advantage of?  It would suck for our dogfooding cluster's IP addresses to be banned by github because someone thinks it's funny to request the status a million times.

Could we do a best-of-both worlds type thing and request the raw data on a daily or hourly basis instead of every query?

bobcatfish: yep that's right!

@sbwsg in this scenario is someone coming into the build cop channel and posting repeatedly? or are they doing this some other way? (i dont think this service would respond to requests that dont come from slack?) would github ban us or would be be temporarily throttled?

if this IS the scenario, im wondering if we'd have a bigger problem - specifically a malicious person in slack

sbwsg: I don't totally agree that a user acting badly in slack is a bigger problem than having our dogfooding cluster rate limited by github.  The fallout from a bad user in slack is an unpleasant slack experience.  The fallout for dogfooding is that all of our PRs no longer merge because the checks we have in dogfooding are blocked.

More generally I'm a little puzzled why we wouldn't write this defensively.

bobcatfish: > The fallout for dogfooding is that all of our PRs no longer merge because the checks we have in dogfooding are blocked.

I'm trying to understand what would happen if we did get rate limited - as far as i can tell the requests this bot makes are unauthenticated, so the limit is pretty low (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) - I'm not sure if we make unauthenticated requests in the rest of our dogfooding (probably we shouldn't given the low rate limiting?), and even if we do, I'm not sure if github would treat these as all coming from the same unauthenticated source:

> Unauthenticated requests are associated with the originating IP address, and not the user making requests.

It seems like nodes in dogfooding have public IPs, so I suppose a node could get rate limited.

But this is only a problem if our CI jobs are also unauthenticated:

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. 

This would mean our CI would also be limited to 60 requests per hour (per node) :O (and actually means we could get locked out by someone opening too many PRs too quickly 😅 )

So long story short:
1. We should probably make sure we are making authenticated requests in our CI
2. If you really want I can make it so only the daily update reads the file but I honestly don't think this is that big of a risk (certainly less of a risk than what our CI is already doing)

PR #712, opened 2021-02-05
Fix the check labels job

afrittoli:
# Changes

The check labels job uses the PR pipeline resource to add a comment
to the PR with instructions for developers in case of failure.
That doesn't work though because in case of failure the sync-up side
of the pipeline resource is not processed, so the commit is not
written.

Fix that by removing the pipeline resource and writing the instructions
to the step log instead. The step log is linked in the github status
details through the tekton dashboard.

In future we might extend the github update service to include
support for comments, but that would require support for results on
failure: https://github.com/tektoncd/pipeline/issues/3749

# Submitter Checklist

review them:

for more details._

/kind misc

### DIFF
--- a/tekton/ci/jobs/tasks.yaml
+++ b/tekton/ci/jobs/tasks.yaml
@@ -148,13 +148,6 @@ metadata:
   description: |
     Verifies that a PR has one valid kind label
 spec:
-  resources:
-    inputs:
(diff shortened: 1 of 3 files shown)


vdemeester (approved):
/lgtm

PR #725, opened 2021-02-12
Add k8s cluster management task to s390x test setup

barthy1:
# Changes

- Add task to create/delete k8s cluster on s390x machine (ssh private key `s390x-k8s-ssh`  is expected)
- Decouple 1 yaml file with s390x tasks to separate files for each task
- Use workspace to store k8s config file
- Add 2h timeout to s390x e2e test pipeline
- extend s390x pipeline cronjob with ssh details

/kind feature
/cc @afrittoli

# Submitter Checklist

review them:

for more details._


tekton/cronjobs/bases/nightly-tests/trigger-nightly-test.yaml
```diff
@@ -64,4 +70,12 @@ spec:
                 value: "s390x"
               - name: KUBECONFIG_SECRET
                 value: "secret"
+              - name: REMOTE_SECRET
```
afrittoli, line 73: NIT: maybe the variables could be named `KUBECONFIG_SECRET_NAME` `REMOTE_SECRET_NAME`

tekton/resources/nightly-tests/bastion-z/cleanup_tekton.yaml
```diff
@@ -0,0 +1,53 @@
...
+kind: Task
+metadata:
+  name: cleanup-tekton-nightly
+spec:
+  params:
+  - name: package
+  - name: resources
+    description: space separated list of resources to be deleted
+    default: "conditions pipelineresources tasks pipelines taskruns pipelineruns"
+  - name: kubeconfig-secret
```
afrittoli, line 11: You are using workspaces for the secret (yay), so you don't need this parameter anymore.
You could instead have a parameter to specify the name of the key within the secret (which is hardcoded to `config` now, in L30 below).

tekton/resources/nightly-tests/bastion-z/deploy_tekton_component.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - name: KO_DOCKER_REPO
+      value: $(params.container-registry)
+    - name: KUBECONFIG
+      value: $(workspaces.k8s-shared.path)/config
+    command:
+    - /bin/bash
+    args:
+    - -ce
+    - |
+      ko apply --platform=linux/$(params.target-arch) -f config/
```
afrittoli, line 38: We could consider here installing our nightly builds instead, so we don't need to rebuild and we test our builds as well. We already have a task to deploy a pre-built tekton, we could then use that too.
Not something for this PR, but we could consider changing it as follow up.

tekton/resources/nightly-tests/bastion-z/cleanup_tekton.yaml
```diff
@@ -0,0 +1,53 @@
...
+    description: secret with kubeconfig for remote cluster
+  resources:
+    inputs:
+    - name: plumbing-source
+      type: git
+    - name: tekton-project-source
+      type: git
+      targetPath: src/$(params.package)
```
afrittoli, line 19: I'm slowly working through the process of removing pipeline resources in various places from our infra - since they're not really supported anymore. I guess this is fine, but eventually we should try and rely on workspaces and the git clone task instead. I don't like having two use a PVC and many pods for that, but that is something that we'll hopefully fix through changes in the runtime model.

afrittoli (commented):
Thank you for this - I have not tested it but it looks ok.
The only thing that I'd ask you to change is to cleanup the unused `kubeconfig-secret` parameter.

tekton/cronjobs/bases/nightly-tests/trigger-nightly-test.yaml
barthy1: updated

tekton/resources/nightly-tests/bastion-z/cleanup_tekton.yaml
barthy1: workspace for ssh key is now used

tekton/resources/nightly-tests/bastion-z/deploy_tekton_component.yaml
barthy1: [e2e pipeline test ](https://github.com/tektoncd/pipeline/blob/f59e1bec5c6b6566370a5dd5f35671a49f577524/test/taskrun_test.go#L125) fails with nightly release (it checks the annotations and expects different value, then nightly release has)

+ testing the latest code instead of run the tests for the same release several times (in case of nightly release delay, for instance) looks better to me, however I am ok with nightly release approach for the future

tekton/resources/nightly-tests/bastion-z/cleanup_tekton.yaml
barthy1: yep, moving to workspaces will be definitely todo for me too

afrittoli (commented):
Thank you!
/approve
/lgtm

PR #731, opened 2021-02-17
Add Catlin to Catalog CI

vinamra28:
# Changes

- Adds a Tekton Task and Pipeline which will run catlin on catalog
  tasks.
- Add corresponding TriggerTemplate.
- Modify EventListener to add catlin jobs.

Signed-off-by: vinamra28 <vinjain@redhat.com>
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/eventlistener.yaml
+++ b/tekton/ci/eventlistener.yaml
@@ -271,3 +271,60 @@ spec:
         - ref: tekton-ci-webhook-issue-labels
       template:
         ref: tekton-community-ci-pipeline
+    - name: catalog-pull-request-ci
+      interceptors:
+        - github:
+            secretRef:
+              secretName: ci-webhook
+              secretKey: secret
+            eventTypes:
+              - pull_request
+        - cel:
+            filter: >-
+              body.repository.full_name == 'tektoncd/catalog' &&
+              body.action in ['opened','synchronize']
+            overlays:
+              - key: git_clone_depth
+                expression: "string(body.pull_request.commits + 1.0)"
+      bindings:
+        - ref: tekton-ci-github-base
+        - ref: tekton-ci-webhook-pull-request
+        - ref: tekton-ci-clone-depth
+        - ref: tekton-ci-webhook-pr-labels
+      template:
+        ref: tekton-catalog-ci-pipeline
+    - name: catalog-comment-ci
+      interceptors:
+        - github:
+            secretRef:
+              secretName: ci-webhook
+              secretKey: secret
+            eventTypes:
+              - issue_comment
+        - cel:
+            filter: >-
+              body.repository.full_name == 'tektoncd/catalog' &&
+              body.action == 'created' &&
+              'pull_request' in body.issue &&
+              body.issue.state == 'open' &&
+              body.comment.body.matches('^/test($| [^ ]*[ ]*$)')
+            overlays:
+            - key: add_pr_body.pull_request_url
+              expression: "body.issue.pull_request.url"
+        - webhook:
+            objectRef:
+              kind: Service
+              name: add-pr-body
+              apiVersion: v1
+              namespace: tektonci
+        - cel:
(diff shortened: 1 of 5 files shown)


tekton/ci/eventlistener.yaml
```diff
@@ -270,4 +270,61 @@ spec:
...
+        - github:
+            secretRef:
+              secretName: ci-webhook
+              secretKey: secret
+            eventTypes:
+              - pull_request
+        - cel:
+            filter: >-
+              body.repository.full_name == 'tektoncd/catalog' &&
+              (body.action == 'opened' || body.action == 'synchronize')
```
bigkevmcd, line 285: One tip here, you can say `body.action in ['opened','synchronize']`

vinamra28:
@afrittoli what else needs to be done apart from this to enable `catlin` job in `tektoncd/catalog`?

tekton/ci/templates/catalog-template.yaml
```diff
@@ -0,0 +1,75 @@
...
+          tekton.dev/pr-number: $(tt.params.pullRequestNumber)
+        annotations:
+          tekton.dev/gitRevision: "$(tt.params.gitRevision)"
+          tekton.dev/gitURL: "$(tt.params.gitRepository)"
+      spec:
+        serviceAccountName: tekton-ci-jobs
+        pipelineRef:
+          name: catlin-linter
+        workspaces:
+          - name: store-changed-files
```
afrittoli, line 50: I don't think you need a PVC to share files between two steps? An emptydir should be enough

vinamra28: :man_facepalming: I forgot to change this. I'll fix this :sweat_smile:

vinamra28:
@afrittoli any updates/feedbacks on this?

tekton/ci/jobs/tekton-catlin-lint.yaml
```diff
@@ -0,0 +1,78 @@
...
+      workingDir: $(resources.inputs.source.path)
+      script: |
+        function detect_new_changed_tasks() {
+          # detect for changes in the task manifest
+          git --no-pager diff --name-only HEAD~$(( $(params.gitCloneDepth) - 1 ))|grep 'task/[^\/]*/[^\/]*/*[^/]*.yaml'|xargs -I {} dirname {}
+        }
+        all_tests=$(detect_new_changed_tasks |sort -u || true)
+        echo -n $all_tests > $(workspaces.store-changed-files.path)/changed-files.txt
+    - name: lint-catalog
+      image: gcr.io/tekton-releases/dogfooding/catlin:latest
```
afrittoli, line 34: It looks good to only validate changed files. 
As a follow-up it might be good to have a way to run the linter on all files too.
If we use a specific sha here, instead of latest, we can detect when the image version is changed and run the linter on all resources in that case.

afrittoli (commented):
Thank you!
This looks good - I have not tested it though. 
If you tested this successfully I'd be happy to merge it.
/lgtm

tekton/ci/jobs/tekton-catlin-lint.yaml
vinamra28: > It looks good to only validate changed files.
> As a follow-up it might be good to have a way to run the linter on all files too.
> If we use a specific sha here, instead of latest, we can detect when the image version is changed and run the linter on all resources in that case.

yeah that would be a good idea. How about adding a nightly trigger pipeline (although I need to take a look how I can make that happen :sweat_smile: ) which will run `catlin` on all the resources?

vinamra28:
> Thank you!
> This looks good - I have not tested it though.
> If you tested this successfully I'd be happy to merge it.
> /lgtm

Yeah I have tested it successfully using my own cluster :smile:

vdemeester (approved):
/meow

PR #732, opened 2021-02-18
Add pipeline to run s390x triggers tests

barthy1:
# Changes

- add corresponding cronjob to trigger s390x tekton triggers tests
- add template to run s390x trigger test pipeline
  - k8s cluster installation
  - pipeline and triggers deployment
  - e2e triggers test
  - k8s cluster removal

/kind misc

# Submitter Checklist

review them:

for more details._


tekton/resources/nightly-tests/triggers-deploy-test-s390x-template.yaml
```diff
@@ -0,0 +1,201 @@
...
+      - name: target-arch
+        value: $(tt.params.targetArch)
+      - name: remote-host
+        value: $(tt.params.remoteHost)
+      - name: remote-port
+        value: $(tt.params.remotePort)
+      - name: remote-user
+        value: $(tt.params.remoteUser)
+      resources:
+      - name: tekton-pipeline-source
```
afrittoli, line 178: I'd rather not continue adding new pipeline resources, but as mentioned earlier I'm ok with it for now.
I'd love to see at some point a follow up that removes them 🙏

afrittoli (commented):
Since the pipeline definition is embedded in the trigger, you could probably manage to have a single trigger and a single pipeline definition in there and reduce duplication a bit... but that might become less readable on the other hand, and pipeline and trigger testing pipeline might also diverge in future, so it's probably better as it is now.

I have not tried this, but it looks ok, thank you!
/lgtm

tekton/resources/nightly-tests/triggers-deploy-test-s390x-template.yaml
barthy1: yeah, I definitely remember about moving to git-clone usage. Could you please install this task to bastion-z namespace? and with the next PR I will use it for all s390x pipelines.

savitaashture: +1 on @afrittoli comment

tekton/resources/nightly-tests/triggers-deploy-test-s390x-template.yaml
```diff
@@ -0,0 +1,201 @@
+# Copyright 2020 The Tekton Authors
```
savitaashture, line 1:
```suggestion
# Copyright 2021 The Tekton Authors
```

barthy1:
@afrittoli @savitaashture thank you for review. I updated year in copyright, as per request. If you are ok that switch to workspace usage will be in next PR, could you please add lgtm and approve this one?

afrittoli (commented):
/approve

barthy1:
And one more question from my side. I started to work already on "workspace instead of PipelineResource" approach and faced that 2 workspaces(in my case first one for source code and second one for k8s config) with 2 PVCs accordingly is a reason of TaskRunValidationFailed failure - "more than one PersistentVolumeClaim is bound". 
What is best practice in this case, just use 1 PVC and 1 workspace for all information which needs to be shared between tasks?

afrittoli:
> And one more question from my side. I started to work already on "workspace instead of PipelineResource" approach and faced that 2 workspaces(in my case first one for source code and second one for k8s config) with 2 PVCs accordingly is a reason of TaskRunValidationFailed failure - "more than one PersistentVolumeClaim is bound".
> What is best practice in this case, just use 1 PVC and 1 workspace for all information which needs to be shared between tasks?

Yes, I would say use one workspace at pipeline level, so you will provision only one PVC.
You can use as many workspaces as you want in your tasks, and you can pass different paths of the pipeline workspace to your task workspaces.

savitaashture:
/lgtm

Thank you @barthy1 :+1:

PR #735, opened 2021-02-23
Only try to fetch team members if team key is present in body

fiunchinho:
/kind feature

# Changes

We want to make sure only members from our organization can trigger pipelines. We don't use the same team schema that you folks are using, and changing the code to work with different team terminologies would be too hard I think. That's why I thought about making the team part optional. With these changes, the interceptor will not fail if the `team` key is not in the body payload. If you want to use it, you can still set it, but if you don't, that's fine.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/interceptors/add-team-members/cmd/add-team-members/main.go
+++ b/tekton/ci/interceptors/add-team-members/cmd/add-team-members/main.go
@@ -102,24 +102,29 @@ func makeAddTeamMembersHandler(orgMembersFetcher, teamMembersFetcher urlToList, 
 			marshalError(err, w)
 			return
 		}
-		// Add the PR body to the original body
+		// Add the org members to the original body
 		jsonBody[RootExtensionsKey].(map[string]interface{})[prExtensionsKey].(map[string]interface{})[orgMembersKey] = orgMembers
-		// Get the team URL from the body
-		teamUrl, err := getTeamUrl(jsonBody)
-		if err != nil {
-			log.Printf("failed to extract the Team URL from the body: %q", err)
-			marshalError(err, w)
-			return
+
+		_, ok := jsonBody[teamKey]
+		if ok {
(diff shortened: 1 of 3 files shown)


tekton/ci/interceptors/add-team-members/cmd/add-team-members/main.go
```diff
@@ -102,24 +102,28 @@ func makeAddTeamMembersHandler(orgMembersFetcher, teamMembersFetcher urlToList,
...
-		}
-		// Get the list of org members from the URL
-		teamMembers, err := teamMembersFetcher(teamUrl, token)
-		if err != nil {
-			log.Printf("failed to get the list of team members: %q", err)
-			marshalError(err, w)
-			return
+
+		_, ok := jsonBody[teamKey]
+		if ok {
```
MarcelMue, line 109: Should we add a log message if this was `ok`? Might make debugging easier in the future. But I know that logs in the current codebase are a bit sparse.

MarcelMue (approved):

tekton/ci/interceptors/add-team-members/cmd/add-team-members/main.go
fiunchinho: Current approach in the codebase seems to be logging only on failure, but I think this could help. Let's see what the `OWNERS` think.

afrittoli: Heh, this interceptor is very simple, I didn't even setup a logging framework.
It's ok to add logging for ok cases too, as long they include info relevant for debugging.

afrittoli (commented):
Thanks for this. The code looks good - would you mind updating the README with the new functionality?

fiunchinho:
Done, please take a look.

afrittoli (commented):
Thanks for the updates
/approve
/lgtm

PR #742, opened 2021-02-25
List all members of the org, not just public members

fiunchinho:
# Changes

We are currently fetching the public members of an organization from the `add-team-members` interceptor. This PR changes `add-team-members` to fetch all members instead.

For that, we change the Github API endpoint from `/public_members` to `/members`. Old behavior can still be achieved if the passed `GITHUB_TOKEN` does not belong to the organization, in which case only public members will be listed. If the user owning the `GITHUB_TOKEN` is in the organization, then [both public and private members will be listed](https://docs.github.com/en/rest/reference/orgs#list-organization-members).

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/interceptors/add-team-members/README.md
+++ b/tekton/ci/interceptors/add-team-members/README.md
@@ -13,14 +13,16 @@ incoming JSON as follows:
 {
   "add_team_members":
   {
-    "org_base_url": "https://api.github.com/repos/tektoncd/",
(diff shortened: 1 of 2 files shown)


afrittoli (commented):
It makes sense, nice change, thank you!
/approve

fiunchinho:
It seems `tide` needs the `lgtm` label to merge this.

PR #757, opened 2021-03-04
Use larger workspaces for s390x pipelines

barthy1:
# Changes

Initial 1Gi size of workspaces was not enough to store source code and do the image builds for s390x testing. Propose is to have 2Gi instead.
Current usage is around 1.2Gi-1.5Gi, but to be on the safe side better to have extra disk space.

/kind bug
/cc @afrittoli 

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
/approve
/lgtm

PR #779, opened 2021-03-14
Fix Catlin to skip deleted tasks

vinamra28:
# Changes

git diff step in Catlin Task was detecting the changed files which
catlin was unable to find.

modified the script which will check for the deleted tasks and if any is task
being removed from the catalog then it will not be tested by catlin.

Signed-off-by: vinamra28 <vinjain@redhat.com>
# Submitter Checklist

review them:

for more details._


PuneetPunamiya:
Just had a question, what if the file is deleted, Catlin would still throw an error saying resource not found in the specified path wdyt ??

vinamra28:
> Just had a question, what if the file is deleted, Catlin would still throw an error saying resource not found in the specified path wdyt ??

if the file is deleted then git diff is going to ignore the deletion with the help of flag `--diff-filter=d`

vdemeester (approved):

PR #780, opened 2021-03-16
Fix namespace specification with kustomize for cleanup RoleBindings

barthy1:
# Changes

Currently the "cleanup old prs/trs" pipeline fails when target namespace is not default one, because there are no permissions to do actions with prs/trs in non-default namespace. The reason is that kustomize applies default namespace value to all resources in the tekton/resources/cd folder and overwrites desired(non-default) namespaces for corresponding RoleBindings which grant the permissions.
Suggested fix is
- set default namespace value to all resources except RoleBindings and ClusterRoleBindings.
- set default namespace only for subject of RoleBindings and ClusterRoleBindings.

/kind bug

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/resources/cd/kustomization.yaml
+++ b/tekton/resources/cd/kustomization.yaml
@@ -1,6 +1,19 @@
-namespace: default
 commonAnnotations:
   managed-by: Tekton
+
+patches:
+- target:
+    kind: ServiceAccount|Deployment|PipelineResource|EventListener|TriggerBinding|TriggerTemplate
+  patch: |-
+    - op: replace
+      path: /metadata/namespace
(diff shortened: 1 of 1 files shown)


afrittoli (commented):
Thanks! 
I was imagining something like an overlay for each target namespace where we want to deploy to, but this works too. If we add more namespaces we'll only have to add roles to the service account, right?
/approve

barthy1:
> I was imagining something like an overlay for each target namespace where we want to deploy to, but this works too. If we add more namespaces we'll only have to add roles to the service account, right?

yes, correct, just new RoleBinding for new namespace

PR #762, opened 2021-03-05
Add pipelines nightly tests for ppc64le architecture

Siddhesh-Ghadi:
This PR is similar to #663. It adds `cronjob`, `tasks` & `service account` to run nightly pipeline tests on remote `ppc64le` cluster.

- Add secret & resource details to docs.
- Update nightly test EventListener to include ppc64le architecture.
- Add TriggerTemplate for nightly pipeline tests on ppc64le.
- Add cronjob to trigger nightly ppc64le pipeline tests.
- Add tasks to deploy, test & clean pipeline components on remote ppc64le cluster.
- Add service account for triggering nightly tests.

The job configurations expect following resources to be present on dogfooding 
- `ppc64le-kubeconfig` used to access ppc64le remote k8 cluster
- `registry-credentials` used to access registry on remote machine
- `registry-certificate` self-signed certificate for registry on remote machine
- `ppc64le-cluster` headless service & endpoint to resolve remote machine address

# Changes

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

/kind feature
/cc @afrittoli

PR #786, opened 2021-03-19
Add nightly tests for cli, triggers & operator on ppc64le architecture

Siddhesh-Ghadi:
# Changes

This PR adds nightly e2e tests on `ppc64le` architecture for `cli`, `triggers` & `operator`. Nightly `pipeline` tests on ppc64le hardware is already enabled via #762.

/kind feature
/cc @afrittoli

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cronjobs/dogfooding/nightly-tests/ppc64le/cli-nightly-test/README.md
+++ b/tekton/cronjobs/dogfooding/nightly-tests/ppc64le/cli-nightly-test/README.md
@@ -1,0 +1,1 @@
+Cron Job to run nightly cli e2e tests on ppc64le hardware.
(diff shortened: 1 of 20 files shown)


vdemeester (approved):
/cc @afrittoli

afrittoli (commented):
I would prefer we did not duplicate common deploy and cleanup script between bastion-z and bastion-p (and more architectures in future. 
These deploy tasks are already duplicating logic which is release pipeline of the respective repo, it will become hard to maintain if we copy it many times.

I guess it's ok to do it for now, but I really would prefer if eventually we used the nightly build to run the tests, which would allow us to know whether the nightly build works or not.

In fact we could get rid of the extra cronjob, and define a trigger that catches the completion of a nightly release to start the tests.

/cc @barthy1 

/lgtm

PR #792, opened 2021-03-24
Add task for nightly release installation, common for all architectures

barthy1:
# Changes

- the arch-independent task for tekton component installation is added(installs latest nightly release of component)
- all pipelines for s390x are rewritten to use this task

/kind misc

# Submitter Checklist

review them:

for more details._


afrittoli:
Sorry for the late review on this!
With triggers split on two release files, this won't work now. 
This [task](https://github.com/tektoncd/plumbing/blob/main/tekton/resources/release/base/install_tekton_release.yaml) already does the install and it works with the latest triggers, perhaps we could use that one instead?

barthy1:
@afrittoli thanks for reviewing this PR!
Actually you are correct, that the current code won't work for triggers :(
However I am not sure if we can use the task you mentioned as is, because it uses resources instead of workspaces.
I doubt, I can port `install-tekton-release` on my own, as it can be complex task with many prs/trs, dependeing on it...
So should I wait till the task you mentioned is ported to work with workspaces or extend proposed task to have extra `deploy-extra-manifest` step?

barthy1:
@afrittoli I've just extended the provided code to support also triggers deployment requirements. Please, review.

afrittoli:
> @afrittoli thanks for reviewing this PR!
> Actually you are correct, that the current code won't work for triggers :(
> However I am not sure if we can use the task you mentioned as is, because it uses resources instead of workspaces.
> I doubt, I can port `install-tekton-release` on my own, as it can be complex task with many prs/trs, dependeing on it...
> So should I wait till the task you mentioned is ported to work with workspaces or extend proposed task to have extra `deploy-extra-manifest` step?

Got it, thank you.
I want to migrate the CD pipelines to not use pipeline resources, but I don't have a good alternative to the cluster pipeline resource yet. I will try to work on it this summer.

tekton/resources/nightly-tests/base/deploy_tekton_component.yaml
```diff
@@ -0,0 +1,34 @@
...
+    env:
+    - name: KUBECONFIG
+      value: $(workspaces.k8s-shared.path)/config
+    command:
+    - /bin/sh
+    args:
+    - -ce
+    - |
+      kubectl apply --filename https://storage.googleapis.com/tekton-releases-nightly/$(params.package)/$(params.version)/release.yaml
+      if [ "$(params.extra-file)" != "" ]; then
```
afrittoli, line 31: I think you'll need to wait for CRD to be established before you install the extra manifest - see https://github.com/tektoncd/plumbing/blob/1ffa936d254ffba2565a248371fbd50ce8f29987/tekton/resources/release/base/install_tekton_release.yaml#L105

barthy1:
@afrittoli sorry for the delay. I've added a command to wait for CRDs.

afrittoli (commented):
/approve
/lgtm

PR #802, opened 2021-04-01
Fix handling of condition tasks in CI

afrittoli:
# Changes

Condition tasks are used to evaluate whether a CI job should be
executed or not. Cloud events generated before a CI task is
executed should be ignored, which is achieved today by using
a specific annotation on the Task "ci.tekton.dev/condition" and
filtering out matching events in the trigger using a CEL filter.

However, because of https://github.com/tektoncd/pipeline/issues/3864,
the annotations are not immediately available in the TaskRun,
which causes the filtering to fail, CI jobs to be marked as started
in GitHub but never completed if the condition was false and thus
no CI job is executed.

Add an extra check based on the name. If the task name starts with
"check-", the event is ignored and no github update is sent.

# Submitter Checklist

review them:

for more details._

/kind bug


vdemeester (approved):
/lgtm

PR #806, opened 2021-04-06
GitHub interceptor - Simplified GitHub configuration.

wlynch:
# Changes

This inteceptor aims to provide simplified configuration for GitHub repository
triggers. It is heavily motivated by the existing
[add-pr-body](https://github.com/tektoncd/plumbing/tree/main/tekton/ci/interceptors/add-pr-body)
and
[add-team-members](https://github.com/tektoncd/plumbing/tree/main/tekton/ci/interceptors/add-team-members)
interceptors, but provides a more opinionated strategy to simplify the config
needed to be set by users.

Things that this does not do:
- Allow approval triggering by label / PR - TODO
- Use authenticated clients / work with private repos - TODO
- PR Status callbacks - demos I've shown rely on https://github.com/tektoncd/experimental/pull/662 for this.

# Submitter Checklist

review them:

for more details._


afrittoli:
Thanks for this! It looks very nice - and I'm happy to have it in plumbing - at least for now.
This could be part of an opinionated set of tools to get up and running quickly with Tekton - like the pipeline as code work @chmouel .

For our dogfooding I'd like to keep a balance between these opinionated kind of efforts and the current "Tekton only" declarative approach. The former are easier to install and configure - the latter allow us to dogfood deeper into Tekton features and have historically allowed us to uncover bugs and need to new features.

My only "concern" with the opinionated approaches is that they tend to take users down a path where adding a feature requires adding a feature to a Tekton based CI requires adding a feature to an extra component and not rely on Tekton or solving the problem in Tekton core. On the plus side this might help keeping Tekton more lightweight, if we agree as a community that we want to maintain opinionated solutions for doing CI on top of Tekton.

tekton/ci/interceptors/github/pkg/github/issue_comment.go
```diff
@@ -0,0 +1,109 @@
...
+	commentCfg := cfg.GetPullRequest().GetComment()
+	if commentCfg == nil {
+		// No approver config - take no action (should be covered by PR handler)
+		return nil, Error(codes.FailedPrecondition, "comment config not enabled")
+	}
+
+	// Check if comment matches.
+	match := commentCfg.GetMatch()
+	if match == "" {
+		match = "/ok-to-test"
```
chmouel, line 43: I think it could be nice if this gets configurable!

chmouel (commented):
Looks really nice @wlynch !

tekton/ci/interceptors/github/README.md
```diff
@@ -0,0 +1,108 @@
...
+triggers. It is heavily motivated by the existing
+[add-pr-body](https://github.com/tektoncd/plumbing/tree/main/tekton/ci/interceptors/add-pr-body)
+and
+[add-team-members](https://github.com/tektoncd/plumbing/tree/main/tekton/ci/interceptors/add-team-members)
+interceptors, but provides a more opinionated strategy to simplify the config
+needed to be set by users.
+
+## Configuration
+
+This interceptor expects it's configuration in an InterceptorParam named
```
mpetason, line 13:
```suggestion
This interceptor expects its configuration in an InterceptorParam named
```

tekton/ci/interceptors/github/README.md
```diff
@@ -0,0 +1,108 @@
...
+                  revision: "main"
+                match: "/ok-to-test"
+```
+
+This is the same as the previous example, but explicitly configures all the
+default fields.
+
+## Extensions
+
+This inteceptor will provide the following extension outputs that can be used in
```
mpetason, line 77:
```suggestion
This interceptor will provide the following extension outputs that can be used in
```

tekton/ci/interceptors/github/README.md
```diff
@@ -0,0 +1,108 @@
+# GitHub Simple Interceptor
+
+This inteceptor aims to provide simplified configuration for GitHub repository
```
mpetason, line 3:
```suggestion
This interceptor aims to provide simplified configuration for GitHub repository
```

mpetason (commented):
small docs based updates

tekton/ci/interceptors/github/README.md
wlynch: Done.

tekton/ci/interceptors/github/README.md
wlynch: Done.

tekton/ci/interceptors/github/README.md
wlynch: Done.

tekton/ci/interceptors/github/config/rbac.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
dibyom, line 2: Does the interceptor actually need all these roles?

tekton/ci/interceptors/github/pkg/github/server.go
```diff
@@ -0,0 +1,108 @@
...
+type Server struct {
+	client *http.Client
+
+	// Maps event types -> Interceptor handlers.
+	router map[string]Interceptor
+
+	// Webhook Secret
+	webhookSecret []byte
+}
+
```
dibyom, line 25: Would be nice to see how much the interceptor SDK can help reduce the boilerplate code here.

tekton/ci/interceptors/github/pkg/github/server_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/http/httptest"
+	"net/http/httputil"
+	"testing"
+
+	pb "github.com/tektoncd/plumbing/tekton/ci/interceptors/github/pkg/proto/v1alpha1/config_go_proto"
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+)
+
+func TestServer(t *testing.T) {
+	//mux := http.NewServeMux()
```
dibyom, line 22: remove commented line?

tekton/ci/interceptors/github/pkg/github/server_test.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+	}
+	b := new(bytes.Buffer)
+	if err := json.NewEncoder(b).Encode(req); err != nil {
+		log.Fatal(err)
+	}
+
+	resp, err := srv.Client().Post(srv.URL, "application-json", b)
+	out, _ := httputil.DumpResponse(resp, true)
+	t.Log(out)
```
dibyom, line 50: Is the log intentional?

dibyom: Should we compare out with a expected response or something?

tekton/ci/interceptors/github/pkg/github/status.go
```diff
@@ -0,0 +1,31 @@
+package github
```
dibyom, line 1: This looks like it would be useful in pkg/interceptors

dibyom: Or as part of the interceptor sdk (cc @jmcshane)

tekton/ci/interceptors/github/pkg/proto/v1alpha1/config.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
```
dibyom, line 2: Just curious on the decision to use proto for the config vs just a go struct?

dibyom (commented):
Looks good. Just a couple of minor issues

tekton/ci/interceptors/github/pkg/github/server.go
wlynch: Absolutely! Once it's ready let me know and I can take a look in a follow up PR. :D

tekton/ci/interceptors/github/pkg/github/server_test.go
wlynch: Done.

tekton/ci/interceptors/github/pkg/github/server_test.go
wlynch: Yes - since t.Log only shows up in stdout if the test fails (or -v is provided), so this is just debug information for failed tests.

tekton/ci/interceptors/github/pkg/proto/v1alpha1/config.proto
wlynch: Primarily a shortcut to codegen structs with getters, json tags, and other common things I didn't want to do by hand. 😅

Even though we're not really using this message for RPCs, it makes it easier to ensure consistent field behavior if/when this changes.

tekton/ci/interceptors/github/config/rbac.yaml
wlynch: Removed. The interceptor wasn't using any of these - I needed them to set up the trigger EL for testing.

tekton/ci/interceptors/github/pkg/github/status.go
jmcshane: yeah, i think we can move everything from pkg/interceptors/sdk up to just pkg/interceptors. I think this means I agree with you that this function may make sense in tree as a status helper there

tekton/ci/interceptors/github/pkg/github/issue_comment.go
```diff
@@ -0,0 +1,109 @@
...
+	path := approverCfg.GetPath()
+	if path == "" {
+		path = "OWNERS"
+	}
+	org := event.GetRepo().GetOwner().GetLogin()
+	repo := event.GetRepo().GetName()
+	number := event.GetIssue().GetNumber()
+	commentAuthor := event.GetComment().GetUser().GetLogin()
+
+	fc, _, _, err := client.Repositories.GetContents(ctx, org, repo, path, &github.RepositoryContentGetOptions{Ref: approverCfg.GetRevision()})
```
chmouel, line 65: Am I understanding right, this gets the OWNERS file from the pull request itself? Shouldn't be from the main branch?  (or as configured in github.DefaultBranch from the pr event)

wlynch: Not quite - this gets the revision to fetch from the interceptor config. If the revision is empty, GitHub uses the repo's default branch.
We avoid using the pull request revision to prevent users from hijacking the config.

PR #829, opened 2021-05-06
Fix job that is supposed to close rotten issues/prs 🥀

bobcatfish:
# Changes

I was looking at some PRs and issues today that have been rotten for a
very long time and I was trying to figure out why they weren't getting
automatically closed.

I think that that query was incorrect and was looking for old issues that
were labelled with BOTH stale and rotten, but what we actually want is
issues that are labelled with rotten only, so I think this will do the
trick!

And then I guess we'll see a flurry of automatic closures?? 😅

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1364,7 +1364,7 @@ periodics:
       - |-
         --query=org:tektoncd
         -label:lifecycle/frozen
-        label:lifecycle/stale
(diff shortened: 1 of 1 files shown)


jerop (approved):
hope not too many get closed 🤞🏾

PR #831, opened 2021-05-10
Fix TEPs table lint

nikhil-thomas:
Add `teps/README.md` as arugment to `git diff --exit-code` command.
So that the only diff in table (README.md) can fail table lint.

needed for PRs like https://github.com/tektoncd/community/pull/315

Signed-off-by: Nikhil Thomas <nikthoma@redhat.com.com>

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/jobs/tekton-teps-validation.yaml
+++ b/tekton/ci/jobs/tekton-teps-validation.yaml
@@ -25,7 +25,7 @@ spec:
   - name: teps-table
     image: alpine/git:latest
     workingDir: $(resources.inputs.source.path)
-    args: ['diff', '--exit-code']
+    args: ['diff', '--exit-code', '$(params.teps-folder)/README.md']
 ---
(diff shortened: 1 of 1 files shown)


nikhil-thomas:
@afrittoli @vdemeester 
needed for prs like https://github.com/tektoncd/community/pull/315

afrittoli:
I don't think this change is needed. 
I agree that there should normally be any diff outside of the `README.md`, but still it's valid to check for it and fail in case any change is found.
An example could be a PR that changes to the teps table script so that it makes changes outside of the README. That should not be allowed, and the job as it is today would catch that and fail.

nikhil-thomas:
> An example could be a PR that changes to the teps table script so that it makes changes outside of the README. That should not be allowed, and the job as it is today would catch that and fail.

that's right 😛 

for some reason i thought, the old README.md was from the tep-lint image, which is build with `main` branch.
closing this pr.

PR #834, opened 2021-05-11
Add ingress overlay for results.

wlynch:
# Changes

This should allow results to have its own DNS name + TLS for serving
results externally.

Largely based off of https://github.com/tektoncd/plumbing/blob/main/tekton/cd/dashboard/overlays/dogfooding/ingress.yaml

I'm not 100% confident that this is correct, so would definitely appreciate a thorough look at this PR. 🙏 

# Submitter Checklist

review them:

- [-] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/tekton/cd/results/overlays/ingress.yaml
+++ b/tekton/cd/results/overlays/ingress.yaml
@@ -1,0 +1,23 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+    acme.cert-manager.io/http01-edit-in-place: "true"
+    cert-manager.io/cluster-issuer: letsencrypt-prod
+    dns.gardener.cloud/dnsnames: 'results.dogfooding.tekton.dev'
+    dns.gardener.cloud/ttl: "3600"
+  name: results
+  namespace: tekton-pipelines
+spec:
+  tls:
+  - secretName: tekton-results-tls
+    hosts:
+    - results.dogfooding.tekton.dev
+  rules:
+  - host: results.dogfooding.tekton.dev
+    http:
+      paths:
+      - backend:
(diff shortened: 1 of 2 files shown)


tekton/cd/results/overlays/ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+  tls:
+  - secretName: tekton-results-tls
+    hosts:
+    - results.dogfooding.tekton.dev
+  rules:
+  - host: results.dogfooding.tekton.dev
+    http:
+      paths:
+      - backend:
+          serviceName: tekton-results-api
```
nikhil-thomas, line 21: it is `tekton-results-api-service`, isn't it. 🧑‍💻

wlynch: Good catch! Fixed.

afrittoli:
Sorry, I was late to review this one.

The ingress looks good, but the overlay should have one folder per environment, like pipeline and other projects:
```
- base
- overlays
      |- dogfooding
      \- robocat
```

It's fine to have only `dogfooding` if that's the only place we deploy to for now.
Overlays are deployed along with a release through the `deploy-tekton-release` trigger template, which is executed:
- on demand, using the [`deploy-release.sh`](https://github.com/tektoncd/plumbing/blob/main/scripts/deploy-release.sh) script, on the `dogfooding` cluster
- nightly, via the `tekton-release-cd-trigger-robotcat-*` cronjobs, on the `robocat` cluster

If you only want the ingress to be synced to the cluster, you need to place it somewhere it will be picked by the hourly job that deploys everything under `tekton/resources`, as specified by the chain of `kustomization.yaml` files.

PR #845, opened 2021-06-03
Results API: Add GCP LB specific annotations + resources.

wlynch:
# Changes

This adds GCP annotions to configure the GCP load balancer for use with gRPC +
a BackendConfig for configuring the load balancer health checks so that
traffic can be successfully routed to the backend.

The BackendConfig uses the prometheus port as a quick and dirty
mechanism to check whether the pod is alive. This can be replaced with a
more comprehensive health check later if needed.

https://medium.com/google-cloud/gke-grpc-ingress-loadbalancing-4b9cdbc09758
was used as a model for many of these configuration settings.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cd/results/overlays/dogfooding/service.yaml
+++ b/tekton/cd/results/overlays/dogfooding/service.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    cloud.google.com/backend-config: '{"default": "tekton-results-api"}'
+    cloud.google.com/neg: '{"ingress": true}'
+    cloud.google.com/app-protocols: '{"grpc":"HTTP2"}'
+  name: tekton-results-api-service
+  namespace: tekton-pipelines
--- a/tekton/cd/results/README.md
+++ b/tekton/cd/results/README.md
@@ -1,0 +1,13 @@
+# Previewing changes
+
+To see the generated config, run:
+
+```sh
+$ kustomize build overlays/dogfooding
+```
+
+To diff the changes against the remote cluster:
+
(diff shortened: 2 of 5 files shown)


tekton/cd/results/overlays/dogfooding/service.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    cloud.google.com/backend-config: '{"default": "tekton-results-api"}'
+    cloud.google.com/neg: '{"ingress": true}'
+    cloud.google.com/app-protocols: '{"grpc":"HTTP2"}'
+  name: tekton-results-api-service
```
bobcatfish, line 8: is it possible to include some comments or docs explaining how this service relates to [this service](https://github.com/tektoncd/results/blob/9c6dd2231d9e05bf55bb371e966e3e98295f2149/config/api.yaml#L72-L88)? (i assume that this config is augmenting the service config from the results repo? im not quite understanding how these fit together 🙏 )

wlynch: Added a small readme to show how to preview/diff the changes, also changed the base to a remote resource to make it clearer where these resources are coming from.

Hopefully this is enough? 🤞  Let me know!

bobcatfish: ah okay, so this is a 'kustomization' on top of a published release.yaml? i think ive got it! thanks for adding the readme

bobcatfish (commented):
/lgtm

PR #871, opened 2021-08-02
Add combine script, to combine two manifest lists and push the result

imjasonh:
# Changes

This will be used to build a combined base image from our current
distroless/static base image, and a Windows base image, to produce a
base image that supports both.

This will, in turn, provide a base image for Tekton images so they can
be built with multi-OS support, in addition to their existing
multi-arch(-but-only-Linux) support today.

Ref https://github.com/tektoncd/pipeline/issues/182

# Submitter Checklist

review them:

- [y] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [y] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._

/kind feature


cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+func main() {
+	src1, src2, dst := os.Args[1], os.Args[2], os.Args[3]
```
wlynch, line 35: We should probably do basic arg validation so this doesn't panic / someone doesn't try to do a 3-way merge. 

Up to you if you want to do the validation yourself or use something like [cobra](https://pkg.go.dev/github.com/spf13/cobra).

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+	log.Println("combining", src1, src2, "into", dst)
+	dstr, err := name.ParseReference(dst)
+	if err != nil {
+		log.Fatal(err)
```
wlynch, line 39: Add flavor so we know what's failing?

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+
+	plats := map[string]bool{}
+	var adds []mutate.IndexAddendum
+	add := func(idx v1.ImageIndex) error {
+		mf, err := idx.IndexManifest()
+		if err != nil {
+			return err
+		}
+		for _, desc := range mf.Manifests {
+			b, _ := json.Marshal(desc.Platform)
```
wlynch, line 66: Check error

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+			if plats[string(b)] {
+				return fmt.Errorf("conflicting platform %+v", *desc.Platform)
+			}
+			plats[string(b)] = true
+			log.Printf("found platform %+v", *desc.Platform)
```
wlynch, line 71: We probably need a nil check for platform - json.Marshal can apparently handle this just fine https://play.golang.org/p/EKbmzQmpuJv

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+		log.Fatal(err)
+	}
+	log.Println("---", src2, "---")
+	if err := add(src2i); err != nil {
+		log.Fatal(err)
+	}
+
+	dsti := mutate.AppendManifests(mutate.IndexMediaType(empty.Index, types.DockerManifestList), adds...)
+	mf, _ := dsti.IndexManifest()
+	b, _ := json.MarshalIndent(mf, "", " ")
```
wlynch, line 95: Check errors

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+	log.Println(string(b))
```
wlynch, line 96: Might be useful to use fmt.Println here to print to stdout instead so this can be piped to other tools.

cmd/combine/main.go
```diff
@@ -0,0 +1,103 @@
...
+			adds = append(adds, mutate.IndexAddendum{
+				Add:        img,
+				Descriptor: desc,
+			})
+		}
+		return nil
+	}
+	log.Println("---", src1, "---")
+	if err := add(src1i); err != nil {
+		log.Fatal(err)
```
wlynch, line 86: Add flavor text

wlynch (commented):
Neat! Just some minor things, but otherwise LGTM!

Bonus points if you can figure out a way to test this while mocking out the remote calls. Maybe pull out the merging into a tested func?

cmd/combine/main.go
imjasonh: This error message will tell us the reference it was unable to parse:

```
2021/08/03 16:46:57 could not parse reference: gcr.iofjfjfjfjf:::LKK
```

cmd/combine/main.go
imjasonh: Done.

cmd/combine/main.go
imjasonh: Let's just reject a nil platform, since it's ambiguous what to do in that case.

cmd/combine/main.go
imjasonh: Done

cmd/combine/main.go
imjasonh: Good call, done.

cmd/combine/main.go
imjasonh: Done

cmd/combine/README.md
```diff
@@ -0,0 +1,65 @@
+# combine manifest lists
+
+This tool combines two Docker manifest lists ("multi-arch images") into one that provides all the platforms supported by both manifest lists.
+
+It fails if both images provide the same platforms, or if either isn't a manifest list.
+
+# demo
+
+```
+$ go run ./ \
```
vdemeester, line 10: Naive question: would we want to be able to support more than 2 images in the future ?

vdemeester (approved):
/test .*

cmd/combine/README.md
imjasonh: Maybe, if we find another architecture or OS we'd like to support that aren't covered by distroless+nanoserver.

Windows platforms are weird too, so we might end up wanting to combine distroless+windows-vX+windows-vY, etc.

I don't think we should support that for now though since for now we only need to combine two images. It should be relatively straightforward to modify this to combine multiple manifests if/when we decide to.

bobcatfish:
Don't mind me, just trying something out:

/run try-out-tekton

bobcatfish:
hahaha well that sure didnt work - sorry for the noise @imjasonh ...  now im not sure how to remove that check 😅

bobcatfish:
I forced it to be successful 🙏  sorry again for the noise

sbwsg:
One more try to bump the yamllint task:

/test plumbing-yamllint

sbwsg:
Let's see if pulling this out the merge queue gets the machinery moving again.

/approve cancel

jerop:
dropping lgtm to unblock tide

/lgtm cancel

jerop:
re-lgtming now that tide is unblocked and plumbing-yamllint has passed 

/lgtm

PR #875, opened 2021-08-13
Update yamllint to not use resources 🤩

bobcatfish:
# Changes

As part of TEP-0074 (https://github.com/tektoncd/community/pull/480) I'm
trying to update some of the plumbing use of PipelineResources to not
use them anymore.

I picked this yamllint pipeline as a simple example; I thought I'd need
the experimental pipeline to taskrun controller, but in this case
conditions are in the mix - when expressions are the way to update those
but they aren't supported in the pipeline to taskrun controller yet.
Also the original version would run 3 separate pods, and each would
separately clone the git resource. So in this updated version, even
though it will run 4 separate pods, it will only clone the git repo once
which seems like an optimization. And once we support when expressions
in the pipeline to taskrun controller (seems like it could be easy
enough? they execute sequentially after all!) this could all happen in
one pod.

Anyway this seems like an okay first step; next I'll tackle something
that can use the experimental pipeline to taskrun controller.

I tested this by setting up my own eventlistener with the same bindings
and trigger template on my own repo and it seems to work - I didn't have
all of the event triggering setup tho so the real test will be live :D

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/jobs/conditions.yaml
+++ b/tekton/ci/jobs/conditions.yaml
@@ -62,3 +62,68 @@ spec:
       REGEX="$(echo $(params.gitHubCommand) | awk '{ print $2}')"
       [[ "$REGEX" == "" ]] && REGEX='.*'
       echo "$(params.checkName)" | grep -E "$REGEX"
+---
+# These tasks replace the above conditions with Tasks that have results which can be used in when expressions
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: check-git-files-changed
+  namespace: tektonci
+  annotations:
+    description: |
+      Returns the number of files (matching the regex) that have changed in the provided git repo
+spec:
+  params:
+    - name: gitCloneDepth
+      description: Number of commits + 1
+    - name: regex
+      description: Regular expression to match files changed
+  workspaces:
+    - name: source
+  results:
+    - name: numFilesChanged
+      description: |
+        Contains the number of files that have changed in the last N commits from
+        the revision of the git repo that is available in the source workspace
+  steps:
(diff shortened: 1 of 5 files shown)


tekton/ci/jobs/conditions.yaml
```diff
@@ -62,3 +62,68 @@ spec:
...
+        the revision of the git repo that is available in the source workspace
+  steps:
+    - name: count-files
+      image: alpine/git
+      script: |
+        set -o pipefail
+
+        BACK="HEAD~$(( $(params.gitCloneDepth) - 1 ))"
+
+        cd $(workspaces.source.path)
```
wlynch, line 96: No tests make me sad, but I'm not sure if there's a better way to do this for something so simple. :(

wlynch (approved):
/lgtm

tekton/ci/jobs/conditions.yaml
bobcatfish: now that we have both params + results (and workspaces i guess XD) it be cool if we could provide some tools that did some simple input output tests 🤔  (ala https://github.com/tektoncd/pipeline/issues/1289) - would be harder for tasks and pipelines with side effects but we could start with the ones that dont and work our way through it (maybe use hermetic mode to suppress side effects? and or use something like workflows to make it easier to isolate the parts with side effects...)

(as a side note when i started working on updating the triggering bits, i realized that a lot of the problems i ran into could have been easily caught by some kind of static analysis - e.g. i initially removed the embedded pipeline resource but forgot to add the required workspace - would have been super cool if my IDE could have just been like "hey you are missing a required workspace here...")

PR #876, opened 2021-08-19
Fix ko-gcloud build

AlanGreene:
# Changes

An `ARG` declared before a `FROM` is outside of a build stage, so it can’t be
used in any instruction after a `FROM`. To use the default value of an `ARG`
declared before the first `FROM` use an `ARG` instruction without a value
inside of a build stage.

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

This meant that we were failing to download and extract the go tarball as the GO_VERSION
was unset:
```
INFO[0731] RUN tar -C /usr/local -xzf go${GO_VERSION}.tar.gz 
INFO[0731] cmd: /bin/sh                                 
INFO[0731] args: [-c tar -C /usr/local -xzf go${GO_VERSION}.tar.gz] 
tar: invalid magic
tar: short read
```

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/images/ko-gcloud/Dockerfile
+++ b/tekton/images/ko-gcloud/Dockerfile
@@ -27,6 +27,7 @@ ENV GO111MODULE on
 RUN go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4
 
 FROM google/cloud-sdk:296.0.0-alpine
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/lgtm

PR #883, opened 2021-09-01
results dogfooding: Use external DNS hostname in watcher.

wlynch:
# Changes

Since the API service is configured with the SSL cert for
`results.dogfooding.tekton.dev`, the watcher is unable to connect to the
API via the internal service hostname. This changes the watcher to use
the external hostname so that the SSL cert validation succeeds and we
can successfully connect.

We cannot terminate TLS at the ingress, since gRPC requires encrypted
connections when used with credentials. A potential alternative would be to set
up a different SSL cert for the internal host, but I'm not sure
that gives us significant added benefit over using the external hostname.

Fixes https://github.com/tektoncd/results/issues/131

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cd/results/overlays/dogfooding/kustomization.yaml
+++ b/tekton/cd/results/overlays/dogfooding/kustomization.yaml
@@ -2,6 +2,7 @@ bases:
 - ../../base
 patchesStrategicMerge:
 - service.yaml
+- watcher.yaml
 resources:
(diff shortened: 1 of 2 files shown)


afrittoli (commented):
/approve

vdemeester (approved):
/lgtm

PR #895, opened 2021-09-24
Build multi-arch go-rest-api-test image

barthy1:
# Changes

`go-rest-api-test` image is used in many Tekton catalog tests. To be able to run the same tests for different architectures, the image should be multi-arch one.
To build the multi-arch image new nightly cronjob is introduced.

Also 2 values for `build type` parameter of multi-arch image build are introduced - `docker` and `ko` to specify which tool to use for the build. From eventlistener side both types are already supported.

/kind feature

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/cronjobs/README.md
+++ b/tekton/cronjobs/README.md
@@ -213,11 +213,13 @@ spec:
               value: tekton/images/myimage
 ```
 
-4. Add PLATFORMS to `cronjob.yaml` if you want to build multi-arch image. List target architectures for the image.
+4. Add PLATFORMS and BUILD_TYPE to `cronjob.yaml` if you want to build multi-arch image. List target architectures for the image. For BUILD_TYPE use [docker](../resources/images/docker-multi-arch-template.yaml) or [ko](../resources/images/ko-multi-arch-template.yaml) value to choose the image build tool.
 ```
 ...
             - name: PLATFORMS
               value: "linux/amd64,linux/s390x,linux/ppc64le"
+            - name: BUILD_TYPE
+              value: docker
 ```
 **Note**
 Please, make sure that the image is buildable for listed architectures. For instance, base image in corresponding Dockerfile should support the same(or larger) list of architectures.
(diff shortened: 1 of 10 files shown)


tekton/cronjobs/README.md
```diff
@@ -213,11 +213,13 @@ spec:
               value: tekton/images/myimage
 ```
 
-4. Add PLATFORMS to `cronjob.yaml` if you want to build multi-arch image. List target architectures for the image.
+4. Add PLATFORMS and BUILD_TYPE to `cronjob.yaml` if you want to build multi-arch image. List target architectures for the image. For BUILD_TYPE use `docker` or `ko` value to choose the image build tool
```
afrittoli, line 216: NIT: missing dot at the end. Perhaps you could add a link to the template used for each of `ko` and `docker`?

tekton/cronjobs/bases/image-build/trigger-image-build.yaml
```diff
@@ -83,4 +84,6 @@ spec:
                 value: "gcr.io/tekton-releases/dogfooding/myimage:latest"
               - name: PLATFORMS
                 value: ""
+              - name: BUILD_TYPE
+                value: ""
```
afrittoli, line 88: NIT: perhaps we could have a default value of "docker" - but maybe it's better to require it to be explicit in the cronjob definition.

afrittoli (commented):
Thanks, it looks good!
/approve

tekton/cronjobs/README.md
barthy1: fixed and added. Thank you

tekton/cronjobs/bases/image-build/trigger-image-build.yaml
barthy1: By default we have single platform build - neither platforms nor buildType value

afrittoli: Ah, right, good point, thank you

afrittoli (commented):
/lgtm

PR #896, opened 2021-09-29
Update ko to v0.9.2

imjasonh:
v0.9.2 fixes a regression introduced in v0.9.1 where the built
entrypoint binary would not be correctly added to PATH. Users who
invoked a ko-built container with cmd:my-app would see failures, while
users who invoked with cmd:/ko-app/my-app would not. This is fixed in
v0.9.2.

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
Thanks!
/approve
/lgtm

Issue #910, opened 2021-10-21
Automate Pull Request Checklist 


jerop:
## Feature Request

We have a submitter checklists in pull requests that enforce with our standards, the one for [tektoncd/pipeline](https://github.com/tektoncd/pipeline) is shown below:

![image](https://user-images.githubusercontent.com/18076400/138368151-8ebf108b-4f5d-49b9-9b47-e6b0796f8ebd.png)

Currently, maintainers and reviewers have to manually check that the submitter checklist has been checked off and verify each item in the list. If the submitter checklist has not been checked off, they need to follow up with the user and request them to follow the standards and check them off. 

We can add automation that checks that the checklist has been checked off. If there's an incomplete item, the check fails and alerts the submitter that they have to fix it before their PR can get merged.

## References
- [Submitter Checklist Discussion](https://groups.google.com/g/tekton-dev/c/LKI4rYDQ9wA/m/Y_0WOfU2AwAJ)
- [Tekton Pipelines Contributor and Reviewer Expectations](https://github.com/tektoncd/community/blob/main/standards.md)

PR #911, opened 2021-10-21
Automate Pull Request Checklist - Check Tasks Completed

jerop:
# Changes

As described in https://github.com/tektoncd/plumbing/issues/910, we have a submitter checklists in pull requests that 
enforce with our standards.

Currently, maintainers and reviewers have to manually check that the
submitter checklist has been checked off and verify each item in the
list. If the submitter checklist has not been checked off, they need
to follow up with the user and request them to follow the standards
and check them off.

In this change, we add automation that checks that the checklist has
been checked off. If there's an incomplete item, the check reports that
there are incomplete tasks but does not fail yet. In a future change, we
will make the check fail if the submitter checklist is not checked off.
We hope that enforcing checking off the submitter checklist will encourage
contributors to look into our standards and apply them to their
contributions.

The implementation is modeled after the check-kind-label job.

# Submitter Checklist

review them:

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

### DIFF
--- a/tekton/ci/jobs/tekton-github-tasks-completed.yaml
+++ b/tekton/ci/jobs/tekton-github-tasks-completed.yaml
@@ -1,0 +1,70 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: github-tasks-completed
+  namespace: tektonci
+  description: |
+    Verifies that a PR has all the tasks completed. A task is considered to be any part of the text that looks like a github markdown checkbox ("- []").
+spec:
+  params:
+    - name: body
+      description: The body of the Pull Request
+  stepTemplate:
+    env:
+      - name: HOME
+        value: /tekton/home
+      - name: PULL_REQUEST_BODY
+        value: $(params.body)
+  steps:
+    - name: install-pyyaml
+      image: python:3-alpine
+      script: |
+        pip install pyyaml --user
+    - name: check-github-tasks-completed
+      image: python:3-alpine
+      script: |
+        #!/usr/bin/env python
+        import sys
+        prBodyText = os.getenv('PULL_REQUEST_BODY')
+        incompleteTasks = prBodyText.count("\n- []")
+        if incompleteTasks > 0:
+          msg = "Error: {} incomplete GitHub Tasks found, expecting no incomplete GitHub Tasks".format(incompleteTasks)
+          print(msg)
+          # TODO (jerop): comment out in a follow up PR
+          # Check failed. Return exit code 1.
+          # sys.exit(1)
+        else:
+          print("Found no incomplete GitHub Tasks)
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: tekton-github-tasks-completed
+  namespace: tektonci
+spec:
+  params:
+    - name: checkName
+      description: The name of the GitHub check that this pipeline is used for
+    - name: gitHubCommand
+      description: The command that was used to trigger testing
+    - name: body
+      description: The body of the Pull Request
+  tasks:
+    - name: check-name-matches
+      taskRef:
+        name: check-name-matches
+      params:
+        - name: gitHubCommand
+          value: $(params.gitHubCommand)
+        - name: checkName
+          value: $(params.checkName)
+    - name: github-tasks-completed
+      taskRef:
+        name: github-tasks-completed
+      when:
+      - input: $(tasks.check-name-matches.results.check)
+        operator: in
+        values: ["passed"]
+      params:
+        - name: body
+          value: $(params.body)
--- a/tekton/ci/README.md
+++ b/tekton/ci/README.md
@@ -343,6 +343,7 @@ spec:
     - ref: tekton-ci-webhook-pull-request
     - ref: tekton-ci-webhook-pr-labels
     - ref: tekton-ci-clone-depth
+    - ref: tekton-ci-webhook-pr-body
   template:
     ref: tekton-plumbing-ci-pipeline
 ```
@@ -366,6 +367,7 @@ spec:
     - ref: tekton-ci-webhook-comment
     - ref: tekton-ci-clone-depth
     - ref: tekton-ci-webhook-issue-labels
+    - ref: tekton-ci-webhook-pr-body
   template:
     ref: tekton-plumbing-ci-pipeline
 ```
--- a/tekton/ci/jobs/kustomization.yaml
+++ b/tekton/ci/jobs/kustomization.yaml
@@ -11,3 +11,4 @@ resources:
 - tekton-kind-label.yaml
 - tekton-golang-lint.yaml
 - tekton-yamllint.yaml
+- tekton-github-tasks-completed.yaml
--- a/tekton/ci/shared/bindings.yaml
+++ b/tekton/ci/shared/bindings.yaml
@@ -85,3 +85,13 @@ spec:
   params:
   - name: labels
     value: $(body.issue.labels)
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: tekton-ci-webhook-pr-body
+  namespace: tektonci
+spec:
+  params:
+    - name: body
+      value: $(body.pull_request.body)
--- a/tekton/ci/shared/template.yaml
+++ b/tekton/ci/shared/template.yaml
@@ -126,3 +126,57 @@ spec:
           value: check-pr-has-kind-label
         - name: gitHubCommand
           value: $(tt.params.gitHubCommand)
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerTemplate
+metadata:
+  name: check-github-tasks-completed
+  annotations:
+    triggers.tekton.dev/old-escape-quotes: "true"
+spec:
+  params:
+  - name: buildUUID
+    description: UUID used to track a CI Pipeline Run in logs
+  - name: package
+    description: org/repo
+  - name: pullRequestNumber
+    description: The pullRequestNumber
+  - name: pullRequestUrl
+    description: The HTML URL for the pull request
+  - name: gitRepository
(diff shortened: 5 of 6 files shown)


jerop:
@afrittoli @bobcatfish @dibyom @vdemeester thank you for the initial reviews on this PR - it is ready for another look

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - name: install-pyyaml
+      image: python:3-alpine
+      script: |
+        pip install pyyaml --user
+    - name: check-github-tasks-completed
+      image: python:3-alpine
+      script: |
+        #!/usr/bin/env python
+        import sys
+        prBodyText = """$(params.body)"""
```
sbwsg, line 26: I'm wondering if this could be tricked into executing arbitrary python code because Tekton is doing direct string interpolation. Not 100% sure but if my PR body is something like this:

```
foo""" + print("~~evil script~~") + """bar
```

I think it might render to:

```
prBodyText = """foo""" + print("~~evil script~~") + """bar"""
```

I think we can potentially work around this by placing `params.body` into an environment variable first on the container and then accessing the env var from the python code:

```
env:
- name: PULL_REQUEST_BODY
  value: $(params.body)
# ...
    script: |
      # ...
      prBodyText = os.getenv('PULL_REQUEST_BODY')
      # ...
```

Not totally sure this is a risk vector but figured I'd mention it. wdyt?

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: github-tasks-completed
+  namespace: tektonci
+  description: |
+    Verifies that a PR has all the tasks completed
```
sbwsg, line 7: Tiny nit: suggest expanding this a bit to describe what we consider a task to be. Something like `Verifies that a PR has all the tasks completed. A task is considered to be any part of the text that looks like a github markdown checkbox ("- []")`.

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,68 @@
...
+      image: python:3-alpine
+      script: |
+        pip install pyyaml --user
+    - name: check-github-tasks-completed
+      image: python:3-alpine
+      script: |
+        #!/usr/bin/env python
+        import sys
+        prBodyText = """$(params.body)"""
+        incompleteTasks = prBodyText.count("- []")
```
sbwsg, line 27: Suggest counting lines that start with these characters (or whitespace + these characters) rather than "any" instance of them because I think they might show up naturally in some kinds of pull request descriptions. Here's a snippet of a PR body I'm imagining that could be caught by this accidentally:

```
I changed the type of the function arguments - []string to []int - in order to...
```

I _think_ (though could be wrong!) that the `- []string` bit would be counted in `incompleteTasks`?

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: HOME
+        value: /tekton/home
+  steps:
+    - name: install-pyyaml
+      image: python:3-alpine
+      script: |
+        pip install pyyaml --user
+    - name: check-github-tasks-completed
+      image: python:3-alpine
+      script: |
```
wlynch, line 23: We should probably break this out into a dedicated `.py` with unit tests.

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,68 @@
...
+      description: The body of the Pull Request
+  stepTemplate:
+    env:
+      - name: HOME
+        value: /tekton/home
+  steps:
+    - name: install-pyyaml
+      image: python:3-alpine
+      script: |
+        pip install pyyaml --user
```
wlynch, line 20: We should use a requirements.txt with hash checking enabled to prevent tampering (e.g. similar to go.mod + go.sum).

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: that's a great suggestion but it seems to me that we haven't done that with any other ci jobs in this repo -- or is there one doing this? maybe this should be part of a larger refactor? wouldn't want this job to be a special case or inconsistent with the rest

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: makes sense, the other ci jobs also do the same thing so will make an issue to update them to use env instead of params directly -- thank you!

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: much clearer description - updated it - thanks :)

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: great idea - updated the check!

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: similar to the comment below, doesn't seem like we handle any other ci jobs like that so maybe that should be part of a larger refactor of the plumbing?

tekton/ci/jobs/tekton-github-tasks-completed.yaml
```diff
@@ -0,0 +1,27 @@
...
+      description: The body of the Pull Request
+  tasks:
+    - name: github-tasks-completed
+      conditions:
+        - conditionRef: "check-name-matches"
+          params:
+            - name: gitHubCommand
+              value: $(params.gitHubCommand)
+            - name: checkName
+              value: $(params.checkName)
```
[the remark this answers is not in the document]
jerop: since the above comments, `ci-workspaces` has been removed and `ci` has been updated -- rebased and updated the implementation in this pr as well (including using `when` expressions)

thanks @bobcatfish!

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: the same was discussed earlier that we need to create an image and address it more broadly outside of this PR

tekton/ci/jobs/tasks.yaml
```diff
@@ -132,3 +132,43 @@ spec:
...
+
+        import sys
+
+        prBodyText = """$(params.body)"""
+
+        incompleteTasks = prBodyText.count("- []")
+
+        if incompleteTasks > 0:
+          msg = "Error: {} incomplete GitHub Tasks found, expecting no incomplete GitHub Tasks".format(incompleteTasks)
+          print(msg)
```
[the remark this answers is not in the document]
jerop: @afrittoli would updating github not be handled using cloud events?

https://github.com/tektoncd/plumbing/tree/main/tekton/ci#setting-up-the-update-of-the-status-check

tekton/ci/jobs/tekton-github-tasks-completed.yaml
jerop: created an issue for this - https://github.com/tektoncd/plumbing/issues/971

sbwsg:
Thanks for creating the env var issue!

/lgtm

tekton/ci/jobs/tekton-github-tasks-completed.yaml
wlynch: I'd say that's true only for jobs that are relying on script - my assumption is a majority of tools we're pulling in by image (e.g. golanglint-ci) have tests in their underlying repo. Ideally we should aim for Tasks to have the same level of testing that we have for the rest of our repos - at the end of the day we're running both in prod eventually 😅 .

This is a good signal that we're shying away from writing tests for our script based Tasks because it's difficult. Similar to the requirements.txt thread below - I suspect that this friction is mostly coming from the need to first create a docker image / the fact that scripts are embedded in yaml files, making them unusable for their language's standard tooling. I think this is a good opportunity for improvement, since I'm sure other Tekton users are falling into the same issues.

Re pushing until a larger refactor - It would be great if we can improve as we go here. Kicking the can down the road is going to create more debt to resolve later, and makes it even more tempting to say "everything else is untested, one more isn't going to hurt". That said, since part of this is looks like a tooling issue with Tekton itself, I'm okay if you want to assign yourself an issue and mark a TODO here to unblock this PR.

jerop: I agree with refactoring out scripts and testing them. Even more, those refactored pieces of software can have their own lifecycle and versioning - this is something I've been thinking about with regards to the Catalog (https://github.com/tektoncd/catalog/issues/29). However, I wouldn't say that `we're shying away from writing tests for our script based Tasks`. It's more so that we don't have a precedence or tooling set up for that yet and I don't have the capacity to set that up right now as part of this PR. I agree with making improvements as we go, but that's practical for small improvements over time and what you're proposing seems like a big change for our tooling. The progressive improvement I'm hoping to make here is automating the pull requests checklist. I've opened an issue https://github.com/tektoncd/plumbing/issues/972 for refactoring scripts in plumbing more broadly, but can't assign myself at this time.

wlynch: SGTM. Thanks! ^_^

sbwsg:
Let's try this again:

/lgtm

lbernick:
lgtm!

bobcatfish:
Thanks for putting this together @jerop ! Very excited to see it in action :D

/approve

PR #936, opened 2021-12-03
Rewrite MarioBot to use Pipelines and Workspaces

### LINKED ISSUE #931: Rewrite Mario bot to use multiple Tasks and Workspaces rather than PipelineResources
### Feature request
The [Mario bot](https://github.com/tektoncd/plumbing/tree/main/tekton/mario-bot) should be rewritten to avoid the use of PipelineResources.

### Use case

This is part of the longer-term plan to [deprecate PipelineResources](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md), which includes moving our dogfooding projects off of them. This bot would then be a good candidate for dogfooding pipeline-to-taskrun (#921).

/assign
/assign bobcatfish

lbernick:
# Changes

PipelineResources are now deprecated, and most of them are replaced with Catalog Tasks.
This commit rewrites the Mario Bot to not use PipelineResources.
The resulting triggers and pipelines have been tested in the dogfooding cluster.
In a subsequent commit, Mario Bot can be used to dogfood the pipeline -> taskrun feature.

Closes #931.

Co-authored-by: Christie Wilson christiewilson@google.com @bobcatfish 
Co-authored-by: Dibyo Mukherjee dibyajyoti@google.com @dibyom 

# Submitter Checklist

review them:

/kind cleanup

### DIFF
--- a/tekton/mario-bot/mario-github-comment.yaml
+++ b/tekton/mario-bot/mario-github-comment.yaml
@@ -5,17 +5,17 @@ metadata:
 spec:
   params:
   - name: pullRequestID
-    value: $(body.taskRun.metadata.labels.mario\.bot/pull-request-id)
-  - name: buildUUID
-    value: $(body.taskRun.metadata.labels.prow\.k8s\.io/build-id)
+    value: $(body.pull-request-id)
   - name: gitURL
-    value: $(body.taskRun.spec.resources.inputs[?(@.name == 'source')].resourceSpec.params[?(@.name == 'url')].value)
+    value: $(body.git-url)
   - name: gitRevision
-    value: $(body.taskRun.spec.resources.inputs[?(@.name == 'source')].resourceSpec.params[?(@.name == 'revision')].value)
-  - name: targetImageResourceName
-    value: $(body.taskRun.spec.resources.outputs[?(@.name == 'image')].resourceRef.name)
+    value: $(body.git-revision)
+  - name: targetImage
+    value: $(body.target-image)
   - name: passedOrFailed
-    value: $(body.taskRun.status.conditions[?(@.type == 'Succeeded')].status)
+    value: $(body.status)
+  - name: buildPipelineRun
+    value: $(body.build-pipelinerun)
 ---
 apiVersion: triggers.tekton.dev/v1alpha1
 kind: EventListener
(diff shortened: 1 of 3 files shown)


dibyom:
While working on this, we found that the clusterrole for this EL was wrong i.e. it was missing access to clusterinterceptors. So, all calls to the EL were failing with failed to resolve clusterinterceptor errors

lbernick:
FYI @afrittoli to fully test this out, the key "token" in the secret "mario-github-token" in the dogfooding cluster will need to be updated to use a valid github token that has permission to comment on plumbing issues.

tekton/mario-bot/mario-github-comment.yaml
```diff
@@ -42,117 +42,154 @@ spec:
...
+      workspace: pr
+    params:
+    - name: mode
+      value: download
+    - name: url
+      value: $(params.gitURL)/pull/$(params.pullRequestID)
+    - name: provider
+      value: github
+    - name: secret-key-ref
+      value: mario-github-token
```
afrittoli, line 108: This seems fine - we also have a custom interceptor available on `dogfooding` that extends the event payload with the whole PR data - which is used for the same use case as here - i.e. that the payload of the comment event in GitHub only includes very little data on the PR itself (the URL mostly).

This is an example of how it is used:
https://github.com/tektoncd/plumbing/blob/def914a109d482355453d7f6dc766e476a371705/tekton/ci-workspace/plumbing/trigger.yaml#L48-L56

The main advantage of the interceptor is that is saves you the extra pod for the task, and it allows to always have the same input params, regardless of the github specific event.

I don't mind using the task, but you might consider switching to the interceptor in a follow-up PR

tekton/mario-bot/mario-image-build-trigger.yaml
```diff
@@ -149,69 +139,126 @@ spec:
...
+      value: $(params.cloudEventSink)
+    - name: eventID
+      value: $(context.taskRun.name)
+    - name: eventType
+      value: "dev.tekton.event.task.$(tasks.build.status)"
+    - name: data
+      value: '{"pull-request-id": "$(params.pullRequestID)", "build-id": "$(params.buildUUID)", "git-url": "https://$(params.gitRepository)", "git-revision": "$(params.gitRevision)", "target-image": "$(params.targetImage)", "status": "$(tasks.build.status)"}'
+    taskRef:
+      name: cloudevent
+      bundle: gcr.io/tekton-releases/catalog/upstream/cloudevent:0.1
```
afrittoli, line 263: This could also be re-written using native cloud-events. 
Those are all sent to a central event listener, so the trigger would have to be added there.
It's nice though to use `finally` in a dogfooding pipeline :)

tekton/mario-bot/mario-github-comment.yaml
```diff
@@ -42,117 +42,154 @@ spec:
...
-                  value: $(tt.params.gitURL)/pull/$(tt.params.pullRequestID)
-              secrets:
-                - fieldName: authToken
-                  secretName: mario-github-token
-                  secretKey: GITHUB_TOKEN
+            
+          comment_template += (
+            '<a href="http://35.222.249.224/?buildid={buildid}&'
+            'namespace=mario">build logs</a>'
+          )
```
afrittoli, line 171: The logs application is not used anymore, we might want to update this with a link to the tekton dashboard: https://dashboard.dogfooding.tekton.dev/#/namespaces/mario/pipelineruns

afrittoli (commented):
Thank you for working on this!

I left a couple of comments, but nothing blocking.
If you want, you could delete the mario bot go app now, as part of this PR, since it won't be used anymore. 

/approve

tekton/mario-bot/mario-github-comment.yaml
lbernick: Added a TODO to use the interceptor.

tekton/mario-bot/mario-image-build-trigger.yaml
lbernick: I think I'll leave as is for now both because there's a lot changing in this PR and to use `finally`, but this is a good idea for the future!

afrittoli:
> > If you want, you could delete the mario bot go app now, as part of this PR, since it won't be used anymore.
> 
> I'm a bit confused-- I can't tell where the go code is being used currently so it's not clear to me why it's OK to delete.

Heh, Mario is rather old and we the limited capabilities in triggers at the time we wrote it as a go app, but that's not needed anymore. I can follow-up with a cleanup PR.

> 
> I got this working though! see [this comment](https://github.com/tektoncd/plumbing/pull/935#issuecomment-987115841)

🎉

afrittoli:
The webhook and notifications to GitHub have been misbehaving a lot today.
I deployed a cloud event broken to get some more visibility in the cloudevents from pipeline, but I need to spend some more time either making this more robust or replacing it with an operator.

afrittoli (commented):
/lgtm

PR #953, opened 2021-12-08
Cleanup legacy CI

afrittoli:
# Changes

Move the last job - yamllint - from legacy to new CI.
The job relied on a local task which used a local container image.
Start using the catalog task instead and remove the container image
definition for yamllint and its corresponding nightly build cronjob.

Anything else in the `tekton/ci` folder is not used anymore: the
ingress, namespace, service account. What I did was:

- cleanup the tekton/ci
- rename tekton/ci-workspaces to tekton/ci
- fix references to tekton/ci-workspace in docs and yamls

Removing the old `tekton/ci` folder means we can also stop relying
on the legacy `tektonci` namespace. There are no more CI resources
deployed in there, with the only exceptions being:

- the custom interceptors
- the github notification jobs

Custom interceptors were deployed manually from the nightly build.
They will be redeployed to tekton-ci. All the references have been
updated to use the tekton-ci namespace.

GitHub notification jobs are triggered by the tekton-events event
listener in the default namespace, and they used to run in tektonci.
That has been changed now to use tekton-ci instead.

There may be other tekton resources outside of plumbing that rely
on the tektonci namespace, so we should not delete it just yet.

# Submitter Checklist

review them:

for more details._

/kind misc


vdemeester (approved):
/lgtm

PR #958, opened 2021-12-09
Dogfood Pipeline to TaskRun with Mario bot

### LINKED ISSUE #921: Dogfood pipeline-to-taskrun in plumbing
The experimental custom task [pipeline-to-taskrun](https://github.com/tektoncd/experimental/tree/main/pipeline-to-taskrun) is a proof of concept for tackling [TEP-0044 Decouple Task Composition from Scheduling](https://github.com/tektoncd/community/blob/main/teps/0044-decouple-task-composition-from-scheduling.md). While it has unit tests (and it "worked on my machine" haha) to increase confidence in this direction (not in using a taskrun to back a pod specifically, but in the concept of being able to do some amount of task composition inside of one pod) we'd like to use this custom task in our dogfooding as well (plus this is a step in [TEP-0074 PipelineResource deprecation](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#proposal) to show the viability of getting PipelineResource like composition without PipelineResources).

This issue is to:
* Install a nightly release of the pipeline-to-taskrun custom controller in the dogfooding cluster (currently done manually, correct me if I'm wrong @afrittoli !)
* Update at least one Pipeline in our plumbing configuration to use this custom task (and/or update a use of PipelineResources to be backed by this custom task instead)

Some potential candidates for some potential PipelineResource uses we could update to use pipeline-to-taskrun:
* https://github.com/tektoncd/plumbing/issues/887
* Looked at the mario bot a bit with @lbernick, the cloud event PipelineResource might be a bit of a challenge (our migration guide doesn't mention it though there is a CloudEvent catalog task that might work)
*  Automation in https://github.com/tektoncd/plumbing/tree/main/tekton/ci-workspace has already been updated to not use PipelineResources but some of these might be good candidates for taking one step further to running in a pod/taskrun

lbernick:
# Changes

Dogfooding Pipeline to Taskrun is part of our strategy for creating a
successor to PipelineResources. This commit rewrites the mario
image builder trigger template to use a PipelineRun containing a
PipelineToTaskRun, plus a finally Task, and rewrites the mario
github trigger template to use a Run of kind PipelineToTaskRun.

PipelineToTaskRun does not support workspace mountPaths or finally Tasks,
so this commit rewrites existing pipelines to avoid using those features
in a PipelineToTaskRun. In addition, PipelineToTaskRun does not support
OCI bundles, so the pull-request and git-clone tasks were installed
in the mario namespace of the dogfooding cluster.

This commit requires updating Triggers in the dogfooding cluster
to pull in [PR: support Runs in TriggerTemplates](https://github.com/tektoncd/triggers/pull/1283).

Closes #921.

# Submitter Checklist

review them:

/kind misc

### DIFF
--- a/tekton/mario-bot/mario-github-comment.yaml
+++ b/tekton/mario-bot/mario-github-comment.yaml
@@ -47,14 +47,16 @@ spec:
   - name: buildPipelineRun
     description: The name of the image build PipelineRun for the logs link
   resourcetemplates:
-  - apiVersion: tekton.dev/v1beta1
-    kind: PipelineRun
+  - apiVersion: tekton.dev/v1alpha1
+    kind: Run
     metadata:
       generateName: mario-comment-github-
     spec:
-      serviceAccountName: mario-listener
-      pipelineRef:
+      ref:
+        apiVersion: tekton.dev/v1alpha1
+        kind: PipelineToTaskRun
         name: mario-comment-github
+      serviceAccountName: mario-listener
       workspaces:
       - name: pr
         volumeClaimTemplate:
@@ -95,7 +97,6 @@ spec:
   - name: download-pr-info
     taskRef:
       name: pull-request
-      bundle: gcr.io/tekton-releases/catalog/upstream/pull-request:0.1
     workspaces:
     - name: pr
       workspace: pr
@@ -168,7 +169,7 @@ spec:
             )
 
           comment_template += (
-            '<a href="http://dashboard.dogfooding.tekton.dev/#/namespaces/mario/pipelineruns/{buildpipelinerun}?pipelineTask=build&step=build-and-push'
+            '<a href="http://dashboard.dogfooding.tekton.dev/#/namespaces/mario/pipelineruns/{buildpipelinerun}?pipelineTask=build&step=build-and-push&'
             'namespace=mario">build logs</a>'
           )
           comment_params['buildpipelinerun'] = '$(params.buildPipelineRun)'
@@ -182,7 +183,6 @@ spec:
     - setup-comment
     taskRef:
       name: pull-request
-      bundle: gcr.io/tekton-releases/catalog/upstream/pull-request:0.1
     params:
     - name: mode
       value: upload
--- a/tekton/mario-bot/mario-image-build-trigger.yaml
+++ b/tekton/mario-bot/mario-image-build-trigger.yaml
@@ -145,8 +145,66 @@ spec:
         mario.bot/pull-request-id: $(tt.params.pullRequestID)
     spec:
       serviceAccountName: mario-releaser
-      pipelineRef:
-        name: clone-and-build
+      pipelineSpec:
+        workspaces:
+        - name: source-code
+        - name: gcp-secret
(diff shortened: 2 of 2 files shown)


lbernick:
Triggers has now been updated to the latest version in dogfooding, so this can now be merged as is.

vdemeester (approved):

tekton/mario-bot/mario-image-build-trigger.yaml
```diff
@@ -145,8 +145,66 @@ spec:
...
+        - name: publish-event
+          params:
+          - name: sink
+            value: $(params.cloudEventSink)
+          - name: eventID
+            value: $(context.taskRun.name)
+          - name: eventType
+            value: "dev.tekton.event.task.$(tasks.clone-and-build.status)"
+          - name: data
+            value: '{"pull-request-id": "$(params.pullRequestID)", "build-pipelinerun": "$(context.pipelineRun.name)", "git-url": "https://$(params.gitRepository)", "git-revision": "$(params.gitRevision)", "target-image": "$(params.targetImage)", "status": "$(tasks.clone-and-build.status)"}'
```
bobcatfish, line 193: that turned out a lot simpler than i expected!

tekton/mario-bot/mario-github-comment.yaml
```diff
@@ -182,7 +183,6 @@ spec:
     - setup-comment
     taskRef:
       name: pull-request
-      bundle: gcr.io/tekton-releases/catalog/upstream/pull-request:0.1
```
bobcatfish, line 185: thanks for explaining this change in the commit message! i saw this line and i was like "im going to recommend explaining why this happened in the commit" AND IT'S ALREADY THERE 🎉 👍 💯  nice!!

tekton/mario-bot/mario-image-build-trigger.yaml
```diff
@@ -145,8 +145,66 @@ spec:
         mario.bot/pull-request-id: $(tt.params.pullRequestID)
     spec:
       serviceAccountName: mario-releaser
-      pipelineRef:
-        name: clone-and-build
+      pipelineSpec:
```
bobcatfish, line 148: pretty much a nit: i was trying to figure out why this creates a PipelineRun that kicks off a custom task vs directly creating a Run (which is done in mario-github-comment.yaml) and im guessing it's b/c the pipeline was defined inline here and the Run doesn't support that - an alternative would be to define the pipeline as a separate type and reference it here, then you don't need the wrapping pipelinerun

anyway nbd!

bobcatfish (commented):
just a nit from me, but overall very cool to see this happening!!

/lgtm

tekton/mario-bot/mario-image-build-trigger.yaml
lbernick: the reason here is actually that PipelineToTaskRun doesn't support finally tasks, so I needed to create a pipeline to wrap the PipelineToTaskRun tasks + the finally task

PR #969, opened 2021-12-15
Reduce the likelyhood of pending CI jobs

afrittoli:
# Changes

Change the name of the notification taskruns, from a generated
one to a fixed one that includes buildUUID and gitHubCheckStatus
only.

The buildUUID is GitHub event ID, so it will be different between
a PR created, PR sync and comment added events, but it will stay
the same throughout the lifecycle of the PipelineRun triggered by
the event.

The gitHubCheckStatus is set to "pending" by all pipeline run
status except for success and fail. This way, once we run the
<buildUUID>-pending taskrun, it won't be possible to run it again
as the name will already be in the cluster.

On a busy cluster it still may happen that the "pending" taskrun
runs after the "pass/fail" one, but this change reduces the
likelyhood.

# Submitter Checklist

review them:

- [-] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

/kind misc

### DIFF
--- a/tekton/resources/ci/github-template.yaml
+++ b/tekton/resources/ci/github-template.yaml
@@ -33,12 +33,15 @@ spec:
   - apiVersion: tekton.dev/v1beta1
     kind: TaskRun
     metadata:
-      generateName: $(tt.params.checkName)-pr-$(tt.params.gitHubCheckStatus)-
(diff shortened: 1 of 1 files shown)


nikhil-thomas:
/approve
✅ lgtm

PR #980, opened 2021-12-28
WIP: Add tep-notifier custom task to comment on PRs which should trigger TEP updates

abayer:
# Changes

Note that this PR does not actually configure any repositories to trigger off the new custom task. It just adds the code for the task, its own build configuration, and a `TriggerBinding` that will be used when actually configuring its usage.

The `tep-notifier` will listen for PR open/edit/close events in configured repositories. When it receives an event, it will check the PR title and body for TEP references, either as `TEP-1234` or as the TEP's URL in the community repository. If it finds TEPs referenced in the PR that aren't recorded as having the appropriate state for the PR's action in the TEP README, it will create a comment (or edit its existing comment, if, say, the PR body has been edited to refer to an additional or different TEP) suggesting that the TEP(s) (with links and titles) should have their statuses updated accordingly.

Currently, for opened/edited PRs, it looks for TEPs in either the `proposed` or `implementable` state to transition to `implementing`, while for closed PRs, it checks if the PR was merged, and if so, looks for TEPs in the `implementing` state to transition to `implemented`.

/kind misc

# Submitter Checklist

review them:

for more details._


lbernick:
Just curious-- what is the reason to create a pipeline with a custom task rather than just normal tasks?

abayer:
This isn't meant to set status on the PR - it's basically the equivalent of a Prow plugin, but built entirely with Tekton. It was suggested that I go this route rather than a Prow plugin, and hey, why not? =)

lbernick:
Sorry, I still don't understand. It sounds like you want to create a pipeline + trigger combo that listens for PR events and comments on the PR? This should be possible with regular tasks, here's an [example](https://github.com/tektoncd/plumbing/blob/main/tekton/mario-bot/mario-github-comment.yaml#L79-L204) I worked on recently.

abayer:
It's definitely possible with a "traditional" pipeline, yeah. But honestly, it's simpler with a custom task - this way, we don't have to pass a secret to a pipeline (any time I can avoid passing secrets around, I take it!), and we don't end up with a rather complicated embedded script to fulfill the logic (i.e., verifying the event's action is one we want to act on, parsing the TEP README, fetching and parsing existing comments on the PR to see if we need to update an existing comment, etc). I did consider just writing the logic in a binary that gets executed by a pipeline, rather than the custom task controller, but I think there's value in dogfooding the custom task functionality here, since it achieves what we want quite well (my initial idea was to write a custom Prow plugin, since I have experience with that, but given that we at least have an aspirational goal of moving away from Prow towards a pure Tekton solution and the value of dogfooding, I went with this approach).

I'm also thinking about using my work on this (once it's, y'know, _done_ and working!) as a starting point for https://github.com/tektoncd/community/blob/main/teps/0071-custom-task-sdk.md. =)

lbernick:
I would expect regular tasks to be easier to maintain, so I'm surprised to hear that a custom task is simpler here. Why would a custom task not need secrets, when a regular task does? Are there any potential improvements to Task that would have made this easier?

I'm sure you have more context than I do on what is easiest, but I would prioritize having less code to maintain over the opportunity to dogfood a custom task. Happy to defer to what you think is best though.

abayer:
re: secrets - I admit that this one is probably just my instincts and not based on anything rational, but I tend to prefer specifying a secret (in this case, on the custom task controller's deployment) in a place where it can't accidentally leak into user-specified code. That isn't a particular issue here, since the same people have control over both the deployment configuration and job definitions, so like I said, that may just be me being me. =)

I'll also admit to personally not being a fan of Python, so I'm always going to default to writing Go over writing Python. That said, I do think that regardless of the implementation language, the logic here is broad/complex enough to be better written as its own thing than as an embedded script. Again, that may be in part personal preference, but my time working on Jenkins Pipeline internals has given me a strong bias against embedding complex logic inside a pipeline/task/etc definition vs inside a separate script/binary/whatever. I'm also a giant nerd for "all my code should be tested", and a script in a task is not the easiest thing to unit test.

I did come at this from the perspective of someone who's spent a lot of time in Prow (and Prow-clone Lighthouse from Jenkins X), so to start, I wasn't thinking of this as "this is a pipeline that needs to be run" so much as "this is something that should fire in response to a webhook event". In my mind, it's more similar to, say, Prow's "[APPROVALNOTIFIER] This PR is NOT APPROVED (etc)" comment than to running unit tests against the PR. We don't ever need to do anything with the source code, etc, meaning that we don't _really_ need a pod for this particular behavior, and that a controller listening for the webhook event and doing its thing would suffice. Philosophically/design-wise, I would tend towards writing this, and for that matter, things like the task that sets GitHub commit statuses, as custom tasks - that approach is faster at runtime, since it doesn't need to spin up a pod, it creates less clutter, etc.

Looking back on discussions on Slack (like https://tektoncd.slack.com/archives/CJ4ERJWAU/p1639598095104300), the "normal task" approach didn't come up. That's probably in part because "oooo, let's dogfood custom tasks!" is more interesting than "just write a script". =) But overall, I do think this is a reasonable direction to go in, and, frankly, having more examples of custom tasks is worth something on its own. I'm not married to this approach, but I am happy with it and think it's worth going with.

lbernick:
Definitely agree with the importance of unit testing and being able to write in your preferred language! Just one final question, have you considered building a Go program into an image and running that in a Task? i.e. I don't think you should have to put your entire script into a Task definition.

abayer:
I honestly didn't consider that initially, which feels more than a little silly in retrospect. =) I'd still lean towards the custom task, though - considering this is going to be triggered for every PR that gets opened/edited/closed in all the project repos, not having to spin up a new pod each time is a real positive, I think. And now that https://github.com/tektoncd/community/issues/594 is developing into shape, this ends up making a ton of sense as part of that broader concept of a custom task that does various TEP-related automation in one place. So I'm going to mark this as WIP for now, since it'll probably get folded into that work.

abayer:
Closing in favor of the broader #983.

PR #981, opened 2022-01-04
Fix two small issues with catlin script param detection

sbwsg:
# Changes

Prior to this commit catlin unit tests would fail because
the expected error message didn't exactly match the output
of the validator.

This commit removes an unused regex pattern and fixes the
unit test.

# Submitter Checklist

review them:


vdemeester (approved):

nikhil-thomas:
@vinamra28

vdemeester (approved):
/lgtm

PR #1055, opened 2022-04-01
Enable nightly builds for Tekton Chains

lcarva:
# Changes

Set up nightly releases for Tekton Chains.


lcarva:
I do not know how to test this locally. I based these changes on what is already setup for Tekton Pipelines. Any pointers would be greatly appreciated!

lcarva:
@afrittoli, thanks for the `/ok-to-test`. I see that the checks still seem to be pending and there's a failure about missing `kind` label. Is there anything I can do from my end to address this?

vdemeester (approved):
/lgtm

PR #1057, opened 2022-04-04
Add GitHub proxy deployment to the prow cluster

afrittoli:
# Changes

Add the deployment of GitHub proxy from k8s test-infra
https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ghproxy.yaml

Identical to the original, except for the node selector, since we
don't have a dedicated node for gh-proxy, at least for now, and the
metrics generation, that has been disabled.

This is not used anywhere for now, but we shall integrate it in
all prow components.

We might want to consider running a single cluster for prow and tekton
in future so they can both benefit from a single in cluster gh-proxy

# Submitter Checklist

review them:

for more details._

/kind feature

### DIFF
--- a/docs/prow.md
+++ b/docs/prow.md
@@ -78,6 +78,10 @@ kubectl apply -f boskos/boskos.yaml # Must be applied first to create the namesp
 kubectl apply -f boskos/boskos-config.yaml
 kubectl apply -f boskos/storage-class.yaml
 
+# Deploy GitHub Proxy
+kubectl apply -f prow/gce-ssd-retain_storageclass.yaml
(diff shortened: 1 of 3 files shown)


afrittoli:
https://github.com/kubernetes/test-infra/tree/master/ghproxy

It's basically a GitHub API cache that can be used with prow components (and perhaps Tekton?) to reduce rate limiting issues with GitHub. I hope at least this to help with peribolos jobs which are very slow... even though this means that we should run ghProxy in the dogfooding cluster as well...

dibyom (approved):

vdemeester (approved):

PR #1061, opened 2022-04-11
Document conditional error with permission script ✍️

bobcatfish:
# Changes

While updating permissions so that @jerop can step in as a governing
board member while I am on leave, I ran into an interruption in the
script execution b/c there was a conditional IAM policy on
tekton-nightly. I traced this back to a conditional policy which gave
@priyawadhwa temporary access (via her Google account at the time) but
had already expired in 2021. After removing this conditional policy the
script ran just fine so I added a brief troubleshooting section to the
script's docs to try to explain this if anyone runs into it in the
future.

# Submitter Checklist

review them:

for more details._


priyawadhwa (approved):
Thanks for fixing it!

PR #1063, opened 2022-04-18
Add basic testgrid config

lbernick:
# Changes

This commit adds a basic testgrid configuration for Tekton's prow instance (under /testgrid),
and configuration for a prow job that runs Configurator (https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/configurator).
Configurator generates a configuration file that is used by Testgrid, based on our prow configuration and the config added to /testgrid,
and uploads it to the tekton-prow bucket in GCS. This job will run whenever our prow configuration is updated.

The goal of this commit is to get the infrastructure that creates a Testgrid configuration file working, so we can set up
a Tekton dashboard in the Kubernetes TestGrid instance. This configuration has been validated by Configurator, but
it's likely that not all of it is correct. However, this change will not break any existing Tekton prow jobs.
The negative consequences are just that our new dashboard tab would not be configured correctly.
This commit adds support only for the prow job "pull-tekton-pipelines-integration-test";
support for other prow jobs will be added in a subsequent commit after the TestGrid infrastructure is set up.

# Submitter Checklist

review them:

- n/a Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -198,7 +198,29 @@ presubmits:
         - "--" # end bootstrap args, scenario args below
         - "--" # end kubernetes_execute_bazel flags (consider following flags as text)
         - "./tekton/images/test-runner/test.sh"
-
+  - name: pull-tekton-plumbing-check-testgrid-config
+    run_if_changed: '^(prow/.*\.yaml)|(testgrid\.yaml)$'
+    decorate: true
+    annotations:
+      testgrid-create-test-group: "false"
+    labels:
+      preset-presubmit-sh: "true"
+    spec:
+      containers:
+      - image: gcr.io/k8s-prow/configurator:v20220412-66078146cd
+        command:
+        - configurator
+        args:
+        - --yaml=testgrid/config.yaml
+        - --default=testgrid/default.yaml
+        - --prow-config=prow/config.yaml
+        - --prow-job-config=prow/config.yaml
+        - --prowjob-url-prefix=https://github.com/tektoncd/plumbing/tree/main/prow/
+        - --update-description
+        - --validate-config-file
+        resources:
+          requests:
+            memory: "1Gi"
   tektoncd/catlin:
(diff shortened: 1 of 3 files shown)


prow/config.yaml
```diff
@@ -303,6 +302,30 @@ presubmits:
           - "--cov-threshold-percentage=0"
           - "--github-token=/etc/github-token/oauth"
 
+  - name: pull-tekton-pipeline-check-testgrid-config
+    run_if_changed: '^(prow/.*\.yaml)|(testgrid\.yaml)$'
```
chases2, line 306: This job is located under "presubmit", meaning it will run before every PR. You probably want it under "postsubmit", where it will run (and update) after a prow or testgrid file is changed.

chases2: The general pattern is to run this image as a presubmit with `--validate-config-file` (to make sure the updates will work) and also run this image as a postsubmit with `--output`

chases2 (changes requested):
The TestGrid config looks great overall. A few suggested tweaks with the prow jobs you're creating, though.

prow/config.yaml
lbernick: got it thanks! I removed "preset-presubmit-sh" from the postsubmit job since it seems to be for presubmits (none of the postsubmits have it) but I think you mentioned that was needed for credentials-- is this OK?

chases2: I'm guessing "preset-postsubmit-sh" might be the equivalent for setting up credentials in a postsubmit: https://github.com/tektoncd/plumbing/blob/0c4e9f3eb7143c105fe41df622f72fda6bed3d1e/prow/config.yaml#L106-L120

lbernick: nice, thanks!

dibyom (approved):

PR #1077, opened 2022-04-29
Remove bad unit test

dibyom:
# Changes

The "cannot fetch url" test case was testing bad http connection errors that
are also tested by other test cases in the same file. In addition, this test
was flaky since sometimes the error message would include the hostname where
the lookup failed.

/kind bug

# Submitter Checklist

review them:

for more details._


afrittoli (approved):
/lgtm

PR #1082, opened 2022-05-12
Add examples test on kind cluster

Yongxuanzhang:
# Changes

This commit adds a new Pipelinerun using `kind-e2e` as the Pipeline,
this Pipelinerun will pass new `e2e-env` and `e2e-script` where users
can define the script to run yaml file examples. Before this commit
examples tests of Tekton Pipeline are in `e2e-tests.sh` and they are
quite flacky. By moving them to a seperate ci job we don't have to rerun
the whole integration test and can save time for developers.

# Submitter Checklist

review them:

for more details._

/kind feature

### DIFF
--- a/tekton/ci/pipeline/template.yaml
+++ b/tekton/ci/pipeline/template.yaml
@@ -80,3 +80,52 @@ spec:
           value: test/e2e-tests.sh
         - name: e2e-env
           value: test/e2e-tests-kind.env
+  - apiVersion: tekton.dev/v1beta1
+    kind: PipelineRun
+    metadata:
+      generateName: pull-pipeline-kind-k8s-v1-21-examples-
+      labels:
+        prow.k8s.io/build-id: $(tt.params.buildUUID)
+        tekton.dev/kind: ci
+        tekton.dev/check-name: pull-pipeline-kind-k8s-v1-21-examples
+        tekton.dev/pr-number: $(tt.params.pullRequestNumber)
+      annotations:
+        tekton.dev/gitRevision: "$(tt.params.gitRevision)"
+        tekton.dev/gitURL: "$(tt.params.gitRepository)"
+    spec:
+      serviceAccountName: tekton-ci-jobs
+      workspaces:
+        - name: sources
+          volumeClaimTemplate:
+            spec:
+              accessModes:
+                - ReadWriteOnce
+              resources:
+                requests:
+                  storage: 1Gi
+        - name: credentials
+          secret:
+            secretName: "release-secret"
+      pipelineRef:
+        name: kind-e2e
+      params:
+        - name: pullRequestNumber
+          value: $(tt.params.pullRequestNumber)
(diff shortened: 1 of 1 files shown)


Yongxuanzhang:
Right now for tekton/pipelines, the examples tests are in alpha and stable integration test. So I guess we either have 2 separate pipelineruns or like this using one but in the script we run both of them

tekton/ci/pipeline/template.yaml
```diff
@@ -80,3 +80,52 @@ spec:
...
+        - name: fileFilterRegex
+          value: '^(cmd/|examples/|images/|pkg/|test/|go\.)'
+        - name: checkName
+          value: pull-pipeline-kind-k8s-v1-21-examples
+        - name: gitHubCommand
+          value: $(tt.params.gitHubCommand)
+        - name: k8s-version
+          value: v1.21.x
+        - name: e2e-script
+          value: test/examples-tests.sh
```
lbernick, line 129: It doesn't look like there's an existing `test/examples-tests.sh` in the pipeline repo. The existing PipelineRun in this trigger template runs the [test/e2e-tests.sh script](https://github.com/tektoncd/pipeline/blob/main/test/e2e-tests.sh) from pipelines. I would suggest either:

- create a new script in pipelines that runs just the examples tests, and pass that script in as a parameter here. Once this is working, remove examples tests from the e2e-tests script.
- OR refactor the e2e-tests script to take parameters for which tests to run, and update the [e2e-kind](https://github.com/tektoncd/plumbing/blob/a749e56d9915c9cd246e60207e1237c63be9ced3/tekton/ci/jobs/e2e-kind.yaml) pipeline to allow passing extra arguments to the test script. Then update the existing e2e PipelineRun in this trigger template to pass a flag to run only integration tests, and update this new one you're adding to pass the flag to run only examples tests.

Same thing for the e2e-env param; there's an existing [`test/e2e-tests-kind.env`](https://github.com/tektoncd/pipeline/blob/main/test/e2e-tests-kind.env) in pipelines but no `test/examples-tests-kind.env`.

lbernick (commented):
Just want to double check whether you've had the chance to test this out on a cluster? One way to do this would be to apply the kind-e2e pipeline to your cluster and then run it using the params you're supplying in this pipelinerun. I'm happy to help with testing!

tekton/ci/pipeline/template.yaml
Yongxuanzhang: Yes we don't have them yet, I was thinking adding this pr first and then add the `examples-tests-kind.env` and `examples-tests.sh` in pipeline repo. Does it work? 
Or I should add them in pipeline first?

Yongxuanzhang: Sure! I will close this pr, test it in my cluster, and if that can work, open a pr to add in pipeline repo first.

PR #1112, opened 2022-07-04
Change the kind-e2e image to use test-runner as its parent

abayer:
# Changes

And update the kind version in test-runner.

This is prep for Prow kind e2e job support, making sure that we've got an image that's got the DinD support we have available in a Prow preset has `setup-kind.sh` copied to `/usr/local/bin/kind-e2e.sh`, without having to copy `setup-kind.sh` to multiple locations in git.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/images/kind-e2e/Dockerfile
+++ b/tekton/images/kind-e2e/Dockerfile
@@ -14,21 +14,8 @@
 
 # Runtime image for common dependencies when running kind tests.
 
-FROM golang:1.17.11-alpine
+FROM gcr.io/tekton-releases/dogfooding/test-runner:latest
 LABEL maintainer "Tekton Authors <tekton-dev@googlegroups.com>"
-
-WORKDIR /kind
-
-RUN GO111MODULE=on go get github.com/google/ko/cmd/ko@v0.6.2
-
-# common util tools
-RUN apk add --no-cache \
-  bash curl docker git jq openssl build-base
-
-# Install kubectl and make sure it's available in the PATH.
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl && chmod +x ./kubectl && mv ./kubectl /bin
-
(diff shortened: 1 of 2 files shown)


abayer:
/assign @afrittoli @vdemeester 

There's a bit of an annoying chicken-and-the-egg problem here, in that I think `kind-e2e` will get built before `test-runner`, since I believe the `TaskLoop` ends up iterating in alphabetical order. I'm not sure what to do about that, and I'm completely open to alternatives. =)

afrittoli:
I didn't setup that job with the `test-runner` image because there's lots of stuff in that image that we don't necessarily need, like gcloud tools, docker in docker setup (not sure we want to run docker on kind on k8s?) and `kubetest`.
But then again maybe it's good to have a single image to maintain.

abayer:
@afrittoli Yeah, I'd prefer not to use `test-runner`, I just feel like setting up a Prow job for kind e2e will be easier if it can just use a single container, rather than a sidecar like the `PipelineRun` approach does, and I _loathe_ duplication. =)

abayer:
@afrittoli Another possibility - I wonder if we could get `tekton/images/kind-e2e/setup-kind.sh` into Pipeline's `vendor/github.com/tektoncd/plumbing` so we could just refer to it directly as the command to run in a job using `test-runner`...I'm guessing that would mean throwing a `dummy.go` in `tekton/images/kind-e2e`...

afrittoli:
> ...yeah, didn't think that would work.

The `/test` command is sent to both `prow` and `tekton` webhooks. 
Since the update of prow though, hook complains about any test job it doesn't know about, which is confusing, but also harmless. Tekton will still receive the test command and process it.

Perhaps we should introduce a `/tekton-test` command instead, but then users will need to be aware which test is Tekton and which is Prow, which is painful.

abayer:
> docker in docker setup (not sure we want to run docker on kind on k8s?)

I'm probably missing something, but how else would we run `kind`?

abayer:
Closing in favor of #1115.

PR #1129, opened 2022-07-06
Add daemonset to Prow cluster to tune sysctl for kind, go back to 3 worker nodes

abayer:
# Changes

https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files seems to be the root of cluster launch issues I've been seeing on the kind clusters. I found https://github.com/kubernetes/test-infra/pull/13515 as a solution to this - I took the current version of the daemonset added in that PR, added a sysctl call to set `fs.inotify.max_user_instances=512` as well, and documented it in docs/prow.md. I've applied this by hand to the Prow cluster already, and it appears to have done the trick.

This also reverts us back to 3 worker nodes per kind cluster, since decreasing the cluster size didn't solve the problem, and I saw some real drop in the kind jobs' speed with 2 nodes.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/docs/prow.md
+++ b/docs/prow.md
@@ -86,6 +86,10 @@ kubectl apply -f prow/ghproxy.yaml
 kubectl apply -f prow/prowjob-schemaless_customresourcedefinition.yaml
 kubectl apply -f prow/prow.yaml
 
+# Deploy daemonset to configure fs.inotify.max_user_[watches,instances] via sysctl.
+# This is to deal with kind having issues like https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files
+kubectl apply -f prow/tune-sysctls_daemonset.yaml
+
 # Create Prow's configuration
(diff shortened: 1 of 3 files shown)


abayer:
@afrittoli fyi, just noticed that the link for `check-pr-has-kind-label` is `https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/$(tt.params.gitHubRepoUnderscore)/1129/check-pr-has-kind-label/1544764738972422144` - note the `$(tt.params.gitHubRepoUnderscore)` - I'm guessing a parameter's missing somewhere in your magic. =)

vdemeester:
> Also makes me wonder if we should have a separate cluster for kind ci tests vs reusing the one that we share with our release builds + hub deployment

I definitely agree with that last part 😛

vdemeester (commented):
/lgtm

PR #1133, opened 2022-07-07
Increase CPU request for most Pipeline jobs to 4000m

abayer:
# Changes

The build and integration tests will all end up using as much CPU as we give them - `ko resolve` being the culprit. So if we're setting a limit of 4000m for those jobs, we should just set the request to 4000m as well.

/kind misc

# Submitter Checklist

review them:

for more details._


vdemeester (approved):
/lgtm

PR #1135, opened 2022-07-08
Experiment with requiring that Pipeline kind jobs can't be on the same node

abayer:
# Changes

I'm trying to figure out why multiple concurrent jobs will sometimes fail a lot of tests at once. This doesn't seem to be CPU or memory contention specifically, since I've seen jobs be just fine when they're sharing the node with other non-kind beefy jobs, so let's try an experiment: make sure each Pipeline kind job is on its own node, to see if the underlying issue is in fact having multiple kind jobs on the same node.

If that hunch does turn out to be true, I'm not entirely sure what the solution will be. But it'll be good to know one way or the other.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1300,6 +1300,7 @@ presubmits:
       preset-presubmit-sh: "true"
       preset-dind-enabled: "true"
       preset-kind-volume-mounts: "true"
+      pipeline-kind-experiment: "true"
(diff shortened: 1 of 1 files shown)


afrittoli (commented):
/approve

vdemeester (approved):
/lgtm

PR #1136, opened 2022-07-12
Try limiting the kind jobs to n2-standard-8-kind nodepool

abayer:
# Changes

These nodes are n2-standard-8s with local SSD ephemeral storage - by requesting 7 cpus, we guarantee that there can only be one job per node, and we're using SSD ephemeral storage to try to improve kind's performance.

This is still a work-in-progress, obviously - if this is the direction we end up going for real, we won't need the `podAntiAffinity`, for example.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1316,6 +1316,13 @@ presubmits:
                   - key: pipeline-kind-experiment
                     operator: Exists
               topologyKey: kubernetes.io/hostname
+      nodeSelector:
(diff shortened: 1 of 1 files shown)


afrittoli (approved):
/lgtm

abayer:
I screwed up the initial version with an unquoted `true`. =)

I'm testing this by hand right now and it's looking good.

vdemeester (approved):
/lgtm

PR #1148, opened 2022-07-15
Add types to parameters in Pipelines in the default namespace

abayer:
# Changes

For some reason, `PipelineRun`s of these `Pipelines`, which are all in the `default` namespace, are hitting validation failures due to the lack of types on parameters. This isn't happening in any other namespace, just `default`. I still can't figure out why this is happening in `default` and nowhere else, but eh, let's fix it for now anyway.

/kind misc

# Submitter Checklist

review them:

for more details._


Yongxuanzhang:
> /approve
> 
> /cc @ywluogg @chuangw6 @Yongxuanzhang not sure if this is a bug yet but something to keep an eye on

Thanks! Is it something similar to previous results problem? Because these existing pipelines don't go through the mutating webhook, but the `Validate()` will be called to check if types is existent.
So we should assume "" equals to `string` type for validation

tekton/resources/cd/catalog-template.yaml
```diff
@@ -57,7 +57,9 @@ spec:
     taskSpec:
       workspaces:
         - name: serviceaccount
+          type: string
```
ywluogg, line 60: This is a bit weird since workspace doesn't have a field Type: https://github.com/tektoncd/pipeline/blob/a0c7d31e23f472cca39690410253a0739f3189db/pkg/apis/pipeline/v1beta1/task_types.go#L122

tekton/resources/images/docker-multi-arch-template.yaml
```diff
@@ -54,9 +61,13 @@ spec:
       taskSpec:
         params:
           - name: contextPath
+            type: string
```
ywluogg, line 64: @Yongxuanzhang interestingly this should be gone through mutation: https://github.com/tektoncd/pipeline/blob/a0c7d31e23f472cca39690410253a0739f3189db/pkg/apis/pipeline/v1beta1/pipeline_defaults.go#L45

abayer:
> Thanks! Is it something similar to previous results problem ([tektoncd/pipeline#5043](https://github.com/tektoncd/pipeline/pull/5043))? Because these existing pipelines don't go through the mutating webhook, but the `Validate()` will be called to check if types is existent.
> So we should assume "" equals to `string` type for validation

It does seem to be very similar, yeah. I couldn't figure out what change to params was equivalent to the results change, or I would have already opened a Pipeline issue. =)

tekton/resources/images/docker-multi-arch-template.yaml
ywluogg: Thanks Andrew for bringing this up! We will be keeping an eye into this

vdemeester (approved):

tekton/resources/images/docker-multi-arch-template.yaml
afrittoli: If I'm not mistaken, resources will only go through mutation if they are created after the mutation logic is deployed on the cluster. Because of that our API / Validation changes we need to account for old resources that may exist on the cluster.  
We may need a new minor release to fix this?

dibyom: One way to deal with this is to call SetDefaults() in the reconciler to ensure event the old resources have the defaults set

jerop:
running into this same issue with param types when trying to run releases -  https://github.com/tektoncd/pipeline/blob/main/tekton/release-pipeline.yaml

saw there's a fix for result types in https://github.com/tektoncd/pipeline/pull/5043 - was there one for param types? can we patch the releases instead of adding the default types?

dibyom:
Won't reapplying the breaking pipelines fix this? (The webhook should apply the default types - this logic has been around for a while I think - I tried with pipelines v0.35 and the default type got added)

abayer:
It doesn't seem to - I fixed everything manually, and then the next time there was a merge in Plumbing, all the pipelines were broken again. Though I guess deleting them and recreating them would work.

afrittoli:
That's maybe  because we use replace and not apply?

abayer:
> That's maybe because we use replace and not apply?

Perhaps? Delete-and-recreate does seem to work, fwiw. I just did that with `catalog-publish` and it got recreated by the cronjob with types.

abayer:
Ok, we had the label `webhooks.knative.dev/exclude: "true"` on the `default` namespace in dogfooding, which meant that even though the pipelines were getting updated regularly, the defaulting never actually happened. Sigh. Everything looks fine now, so I'm closing this.

PR #1175, opened 2022-08-10
Fix: determine whether the destination path already exists before clone

yuzp1996:
# Changes

If directory ${GOPATH}/src/github.com/google/licenseclassifier exists and is not empty we should skip clone again or will cause an error.

Related Issue: https://github.com/tektoncd/pipeline/issues/5193

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/images/test-runner/entrypoint.sh
+++ b/tekton/images/test-runner/entrypoint.sh
@@ -17,8 +17,14 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# Workaround for https://github.com/google/licenseclassifier/issues/20
-git clone https://github.com/google/licenseclassifier ${GOPATH}/src/github.com/google/licenseclassifier
+DIR="${GOPATH}/src/github.com/google/licenseclassifier"
+
+if [ -d "$DIR" ]; then
+    echo "$DIR is not Empty. Skip clone..."
+else
+    # Workaround for https://github.com/google/licenseclassifier/issues/20
+    git clone https://github.com/google/licenseclassifier $DIR
+fi
 
 # actually start bootstrap and the job, under the runner (which handles dind etc.)
 /usr/local/bin/runner.sh "$@"


yuzp1996:
related Issue: https://github.com/tektoncd/pipeline/issues/5193

yuzp1996:
I made a change in https://github.com/tektoncd/plumbing/pull/1175 and maybe it can resolve the problem when running the e2e-upgrade-test.sh but I am not sure how to test this change.

I find that the file entrypoint.sh will be referenced here and will be copied to the image test-runner which is used to run the e2e upgrade test. 
image

Looks like I need first build the test-runner image and replace the container image which runs the e2e upgrade test in the Prow job. 

image

Is there any other test cluster to test this change? I am not sure the step to test this change.

lbernick:
Thank you @yuzp1996!
I'm not sure exactly how to test this but here's what I'd try (maybe @afrittoli would have some more suggestions?)
First build the docker image that includes this file:
`docker build tekton/images/test-runner -t test-runner`
Then try running the unit tests with your newly built image and see if you still run into the licenses issue:
`docker run -v <path-to-your-pipelines-clone>:/pipeline test-runner /pipeline/test/presubmit-tests.sh --unit-tests`

(I'm suggesting unit tests here because they also run with the test-runner image, but don't require creating a k8s cluster. I haven't tested these commands because I unfortunately don't have a working docker setup right now :( I'm pretty sure the other args for the prow job are args for prow, not for tests.)

If this works we'll have to update the [prow job to use a newer version of this image](https://github.com/tektoncd/plumbing/blob/5dbf210387217e509550ce1c522f11a854058d10/prow/config.yaml#L1844) once it's built. It looks like the version of test-runner it's using isn't what the rest of our CI uses and I'm wondering if just updating the image would help.

abayer:
/approve

Yeah, @lbernick's suggestion should be a good way to verify the behavior. That said, I'm pretty sure it's good. =) Once it's merged and new images get built overnight, I'll update the Prow config to use the new tag.

tekton/images/test-runner/entrypoint.sh
```diff
@@ -17,8 +17,14 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# Workaround for https://github.com/google/licenseclassifier/issues/20
-git clone https://github.com/google/licenseclassifier ${GOPATH}/src/github.com/google/licenseclassifier
+DIR="${GOPATH}/src/github.com/google/licenseclassifier"
+
+if [ "$(ls -A $DIR)" ]; then
```
abayer, line 22: Actually, I'd change this line to
```
if [ -d "$DIR" ]; then
```

abayer (changes requested):

abayer:
fwiw, I've done the local test @lbernick described and it needs a bit of a tweak to `docker run -v <path-to-repo>:/home/prow/go/src/github.com/tektoncd/pipeline -w /home/prow/go/src/github.com/tektoncd/pipeline test-runner -- ./test/presubmit-tests.sh --unit-tests`.

When I run that, it works - that's what prompted me to suggest changing to `-d`, since using `ls` results in an unnecessary `ls: cannot access '/home/prow/go/src/github.com/google/licenseclassifier': No such file or directory` showing up in the logs. =)

yuzp1996:
Thanks for your review @lbernick @abayer !

Following @abayer 's suggestion, I changed the way to determine if the target path exists.

yuzp1996:
@abayer  Sorry, I changed the git commit message and pushed it to the repository want to rerun one of the checks but caused the label lgtm to be removed! Can you add a lgtm label when you have time? Thanks!

Issue #483, opened 2020-07-16
Tekton CI should comment on how to re-run tests


afrittoli:
# Expected Behavior
Similar to what prow does, when test job fail they should report on how to re-run them. 
The comment should only be posted once (old one can be deleted), and should include all checks managed by Tekton.
Certain checks (like the kind label check) may want to display extra text.

# Actual Behavior
No comment is displayed.

# Additional Info

PR #1185, opened 2022-08-22
Add new pr-commenter custom task

abayer:
# Changes

Part of #483

This custom task is meant to be used alongside PR status report updates to add/update/delete a comment equivalent to the Prow one listing job failures with links to logs and the command to re-test.

I initially made an attempt to update the `github-add-comment` task in the catalog to do this (https://github.com/tektoncd/catalog/pull/1051) but decided that a custom task frankly made a lot more sense. Its behavior is a pretty direct emulation of how Prow and Lighthouse do this, looking for existing comments by the bot user with a given tag in the comment, and updating/deleting that comment if it exists, creating a new comment otherwise. It's a bit simplified, and instead of using `ProwJob`s to know what should still be in the comment and what should be removed, it specifically just adds or removes the specific job it's reporting on while preserving whatever else was there.

/kind feature

# Submitter Checklist

review them:

for more details._


abayer:
Ok, turns out getting an integration test set up properly is more hassle than it's worth, since this exact same behavior with extremely similar code is covered by Lighthouse's end-to-end tests, so I'm just going to do a manual test locally and then get rid of the WIP.

abayer:
> Shouldn't there be a CRD that defines `PRCommenter` ?

Originally there was, but then I realized that was unnecessary since it's just parameters, but I kept the struct in the same form anyway. 🤷

abayer:
^^ is me testing the bot.

tekton/resources/nightly-release/overlays/pr-commenter/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
+namePrefix: pr-commenter-
+bases:
+  - ../../base
+patchesJson6902:
+  - target:
+      group: triggers.tekton.dev
+      version: v1alpha1
```
afrittoli, line 7: Heh, I need to switch the bases to beta eventually

afrittoli (commented):
This looks great, thank you!
I would love to see a README with an example `Run`, but otherwise it seems ok, including the release resources and cronjob.
/approve

abayer:
> I would love to see a README with an example `Run`, but otherwise it seems ok, including the release resources and cronjob.

I knew I missed something! =) Adding that now.

afrittoli (commented):
Let's get the ball rolling!
Hopefully we can get a nightly and deploy to dogfooding!
/lgtm

PR #1191, opened 2022-08-30
Move Dashboard integration tests from GKE to Kind

AlanGreene:
# Changes

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -808,29 +808,47 @@ presubmits:
   - name: pull-tekton-dashboard-integration-tests
     labels:
       preset-presubmit-sh: "true"
+      preset-dind-enabled: "true"
+      preset-kind-volume-mounts: "true"
     agent: kubernetes
     always_run: true
     decorate: true
     rerun_command: "/test pull-tekton-dashboard-integration-tests"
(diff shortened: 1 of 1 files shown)


AlanGreene:
@tektoncd/plumbing-maintainers please review

abayer:
/approve

Looks good! Hopefully my migration doc was accurate and this just works. =)

AlanGreene:
/hold
hrmm, digging into this a bit more I think it will actually fail at https://github.com/tektoncd/plumbing/blob/5d5152b72f966ad0fda5119ea28ba2ac03cce8bc/tekton/images/test-runner/setup-kind.sh#L314 as we're passing `"./test/presubmit-tests.sh --integration-tests"` which is not a valid command.

@abayer Any objections to removing the surrounding quotes in the script? The only down side I see is that folks using paths containing spaces would need to ensure they're properly escaped in prow/config.yaml rather than relying on the script handling it.

abayer:
Aaaaah, you've got pre-integration stuff to run in `presubmit-tests.sh` - Pipeline doesn't, so we can just call `./test/e2e-tests.sh` directly. Yeah, I think a wrapper script would probably be cleanest.

AlanGreene:
Should be good to go now 🤞
/hold cancel

PR #1211, opened 2022-09-13
Remove version tag labels from pr-commenter controller

abayer:
# Changes

`VERSION_TAG` isn't set in the right task, and I'm being lazy.

/kind misc

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

PR #1212, opened 2022-09-13
Switch `${VERSION_TAG}` to `$(params.versionTag)` in publish tasks

abayer:
# Changes

Since the sed calls had `${VERSION_TAG}` in single-quotes, it was being used literally, not for its value. In Pipeline's `tekton/publish.yaml`, we instead just use `$(params.versionTag)` directly, which will get properly replaced, so let's do that here.

Replaces #1211, since this is the right way to actually do it.

/kind misc

# Submitter Checklist

review them:

for more details._


abayer:
👍 I’ll switch to that.

PR #1231, opened 2022-10-04
Turn pr-commenter back on, using existing params

abayer:
# Changes

As it turns out, the problem wasn't the params stuff but that I misformatted the `PipelineRun` in the trigger template. But hey, I still think this is better than needing new param values, so tada.

/kind misc

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
/approve
/lgtm

PR #1236, opened 2022-10-04
Log errors with SCM client in pr-status-updater

abayer:
# Changes

In dogfooding, it's failing with a "not found" error (which can mean any number of things), but it's working for me locally. I've made a small tweak to the deployment, in case not having `GIT_USER` is causing problems (I don't think it is, but this change matches what's in the configuration for `pr-commenter`), and added some debug logging to better determine what it's actually doing/failing on.

EDIT: Turns out the failure was due to an old version of `pr-status-updater` being installed. Oops. Well, I still think this is worth adding for logging purposes.

/kind misc

# Submitter Checklist

review them:

for more details._


abayer:
/assign @afrittoli 

FYI. =)

wlynch (approved):
/lgtm

PR #1243, opened 2022-10-17
Don't delete/create identical PR comments

abayer:
# Changes

This doesn't fix whatever the underlying problem (i.e., events being processed more than once, causing the triggers for a specific `TaskRun` failure to fire periodically) is, and can still result in a single outdated comment being added for a stale event, but it should at least prevent that same single outdated comment from being deleted and recreated over and over again.

/kind misc

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
Thank you!
/approve

PR #1244, opened 2022-10-17
Fix publish.yaml for pr-commenter

abayer:
# Changes

The `VERSION_TAG` stuff doesn't work for some reason, and is not necessary anyway, since `params.versionTag` is available. So let's get this in sync with what we're doing in pr-status-updater.

/kind misc

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
The ${VERSION_TAG} is within single quotes, you'll need to close the single quotes first.
See my comment on https://github.com/tektoncd/plumbing/pull/1212 - we should not pass parameters directly in the script ideally.

abayer:
In the pr-status-updater work, I hit this too, and switching to double quotes didn't fix it for some reason. I'll try again later and leave this as is for now.

afrittoli:
> In the pr-status-updater work, I hit this too, and switching to double quotes didn't fix it for some reason. I'll try again later and leave this as is for now.

Switching to double quotes does not work, you need single quotes and close them before the variable and the open them again after the variable.

The example in https://github.com/tektoncd/plumbing/pull/1212#issuecomment-1256036946 works on bash/zsh

afrittoli:
@abayer are you still working on this one?

dibyom (approved):

afrittoli:
/hold

@abayer @dibyom please do not merge this as it is, we don't want to use params in scripts directly for security reasons.
I will propose a PR update that uses env variables instead

abayer:
I don't think I can lgtm this, since it started as a PR of mine. =) But it looks fine.

abayer:
Oh, worth mentioning that `pr-status-updater` is using `params.versionTag` currently as well, so that should probably be switched too?

Issue #1239, opened 2022-10-05
Migrate go coverage task to Tekton


chitrangpatel:
`go coverage task` is run by prow. In the interest of dogfooding, we should migrate this to run as a `Tekton pipeline`. The outcome should be the same regardless of whether we run the job on the prow cluster on using Tekton ie.
- We need a PR comment with the coverage report on GitHub.
- We need to upload the artifacts to the tekton-prow bucket so that it can be picked up by Prow's Spyglass.
- We need to update the status of the task on Github in case of success or failure.

To that effect, here is a draft plan that will keep evolving as I learn the details more:

- [x] Write a go coverage Tekton task to run the coverage in Tekton.
- [x] Ensure that the coverage report is sent to the PR on GitHub.
- [x] Upload the artifacts and logs to the tekton-prow GCS bucket.
- [ ] Deploy on the pipelines repository.
- [ ] Compare coverage reports with ones generate by the job running on the prow cluster.
- [ ] Stop using the prow jobs and transition over completely.

afrittoli:
https://github.com/tektoncd/plumbing/pull/1249

chitrangpatel:
This can now be closed. The jobs on the dogfooding cluster seem to be running fine and when clicking on details, it takes users to the tekton dashboard with the logs etc.

PR #1249, opened 2022-10-20
Go coverage Tekton task

chitrangpatel:
This task uses the same knative coverage tool
that prow jobs used. The tool runs the coverage
and sends a comment to github on a PR in a comment.

This is a first PR towards [migrating go coverage task](https://github.com/tektoncd/plumbing/issues/1239) from prow to Tekton for dogfooding purposes.

# Changes

# Submitter Checklist

review them:

for more details._


chitrangpatel:
Using this tool, I was able to send the coverage report to GH in a comment as myself (since I used my personal access token).
Screen Shot 2022-10-20 at 10 32 00 AM

tekton/ci/shared/go_coverage_task.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: go-coverage
+  annotations:
+    description: |
+      Clones a repository from github into an empty-dir workspace.
```
jerop, line 7: this description needs to be updated - add details about go coverage

jerop (commented):
it seems this task could be useful for the community, can we consider adding it to tektoncd/catalog and using it from there?

tekton/ci/shared/go_coverage_task.yaml
chitrangpatel: Thanks! I'm updating it.

chitrangpatel:
> it seems this task could be useful for the community, can we consider adding it to tektoncd/catalog and using it from there?

Sure, I think so. Should I make it a bit more mature here and then move it to tektoncd/catalog?

tekton/ci/shared/go_coverage_task.yaml
```diff
@@ -0,0 +1,71 @@
...
+        - "--artifacts=$(workspaces.shared-workspace.path)/artifacts"
+        - "--profile-name=coverage_profile.txt"
+        - "--cov-target=."
+        - "--cov-threshold-percentage=0"
+        - "--github-token=$(workspaces.github-token.path)/token"
+      workingDir: "$(workspaces.shared-workspace.path)/$(params.repoName)"
+  workspaces:
+  - name: shared-workspace
+    mountPath: /go/src/github.com/tektoncd
+  - name: github-token
```
dibyom, line 71: Would be useful to provide a description of the workspace and where it expects the token to be provided

dibyom:
> Sure, I think so. Should I make it a bit more mature here and then move it to tektoncd/catalog?

Agree, also maybe a more generic task that just does coverage (instead of also uploading + posting to GH might be better?)

tekton/ci/shared/go_coverage_task.yaml
```diff
@@ -0,0 +1,71 @@
...
+        - name: REPO_NAME
+          value: $(params.repoName)
+        - name: REPO_OWNER
+          value: $(params.repoOwner)
+        - name: JOB_NAME
+          value: $(params.jobName)
+      command:
+        - /coverage
+      args:
+        - "--postsubmit-gcs-bucket=tekton-prow"
```
dibyom, line 60: How does the auth for the gcs bucket work? Does the SA for the run have to have permissions to post to this bucket?

tekton/ci/shared/go_coverage_task.yaml
chitrangpatel: Yes. And for others too. I just pushed the descriptions.

tekton/ci/shared/go_coverage_task.yaml
chitrangpatel: I think the gcs bucket is public so anyone has read permissions. eg. I was always able to do something like `gustil ls gs://tekton-prow`

chitrangpatel: The coverage tool fetches the `post-submit` run. It does not write to it. But let me double check this.

chitrangpatel: Interesting. When I open a docker container on my machine and try to run the coverage command, I get this error:
```
2022/10/20 19:37:46 Running workflow: presubmit
2022/10/20 19:37:46 Failed to create client: dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

But, if I simply launch the task on my cluster on GKE, I don't see that. Is it possible that because I'm running the cluster on GKE, my credentials are some how available?

dibyom: yeah, it's using ADC, so it seems tied to Google: https://cloud.google.com/docs/authentication#adc

bendory (commented):
It makes me sad that we don't lean on the existing [golang test](https://hub.tekton.dev/tekton/task/golang-test) task in the Catalog. Suggest opening an issue to nudge us in this direction in the future? (I don't think I want to block this PR for that purpose.)

chitrangpatel:
> It makes me sad that we don't lean on the existing [golang test](https://hub.tekton.dev/tekton/task/golang-test) task in the Catalog. Suggest opening an issue to nudge us in this direction in the future? (I don't think I want to block this PR for that purpose.)

We want to move away from the existing k-native coverage tool that we are using because k-native dropped support for it a while ago. In the docker image, we actually check out an older commit to make the binary. 

However, we do need to then also try and replicate things that the current tool does which is to Identify files affected by the PR, changes in the coverage and display it in a nice way on GitHub. I believe we have some of the pieces but not all of them. Therefore, for now, I simply tried to migrate coverage to Tekton but keeping the same tool.

bendory:
> > It makes me sad that we don't lean on the existing [golang test](https://hub.tekton.dev/tekton/task/golang-test) task in the Catalog. Suggest opening an issue to nudge us in this direction in the future? (I don't think I want to block this PR for that purpose.)
> 
> Completely agree. We want to move away from the existing k-native coverage tool that we are using because k-native dropped support for it a while ago. In the docker image, we actually check out an older commit to make the binary.
> 
> However, we do need to then also try and replicate things that the current tool does which is to Identify files affected by the PR, changes in the coverage and display it in a nice way on GitHub. I believe we have some of the pieces but not all of them. Therefore, for now, I simply tried to migrate coverage to Tekton but keeping the same tool.

Perfect, that's exactly aligned with my thinking -- migrate first, then evolve.

tekton/ci/repos/pipeline/template.yaml
```diff
@@ -51,3 +51,53 @@
...
         - name: e2e-env
           value: test/e2e-tests-kind.env
+  - apiVersion: tekton.dev/v1beta1
+    kind: PipelineRun
+    metadata:
+      generateName: run-go-coverage-
+      labels:
+        prow.k8s.io/build-id: $(tt.params.buildUUID)
+        tekton.dev/source-event-id: $(tt.params.sourceEventId)
+        tekton.dev/check-name: pull-tekton-pipeline-go-coverage
```
afrittoli, line 61: The name is already taken by prow, so we should use a slightly different name here

abayer: and I've just done a quick check, and I'm pretty sure `tekton.dev/check-name` is the value that is used for the GitHub check name by the existing automation.

chitrangpatel: I changed it to `pull-tekton-pipeline-go-coverage-using-tekton` to indicate that the job was run using the Tekton dogfooding cluster. I'm happy to change it to something else too.

tekton/ci/shared/go_coverage.yaml
```diff
@@ -0,0 +1,215 @@
...
+      value "your access token".
+  steps:
+    - name: clone-repo
+      image: alpine/git
+      script: |
+        #!/bin/sh
+        git clone https://github.com/$(params.repoOwner)/$(params.repoName).git
+        cd $(params.repoName)
+        git fetch origin pull/$(params.pullRequestNumber)/head:$(params.gitRevision)
+        git checkout $(params.gitRevision)
```
afrittoli, line 55: Every other CI job uses:

```yaml
    - name: clone-repo
      taskRef:
        name: git-batch-merge
        bundle: gcr.io/tekton-releases/catalog/upstream/git-batch-merge:0.2
```

Any reason for not using it here and having a custom cloning logic?

tekton/ci/shared/go_coverage.yaml
```diff
@@ -0,0 +1,215 @@
...
+      workspaces:
+        - name: source
+          workspace: source
+        - name: github-token
+          workspace: github-token
+    - name: upload-artifacts
+      runAfter: ['run-go-coverage']
+      taskRef:
+        name: gcs-upload
+        bundle: gcr.io/tekton-releases/catalog/upstream/gcs-upload:0.1
```
afrittoli, line 205: There is version 0.3 of this in the catalog - any reason for using 0.1?
Version 0.3 default behaviour is safer https://github.com/tektoncd/catalog/tree/main/task/gcs-upload/0.3#parameters

tekton/ci/shared/go_coverage.yaml
```diff
@@ -0,0 +1,215 @@
+apiVersion: tekton.dev/v1beta1
```
afrittoli, line 1: Other CI jobs are in tekton/ci/jobs

tekton/ci/shared/go_coverage.yaml
```diff
@@ -0,0 +1,215 @@
...
+      image: alpine/git
+      script: |
+        #!/bin/sh
+        git clone https://github.com/$(params.repoOwner)/$(params.repoName).git
+        cd $(params.repoName)
+        git fetch origin pull/$(params.pullRequestNumber)/head:$(params.gitRevision)
+        git checkout $(params.gitRevision)
+      workingDir: "$(workspaces.source.path)"
+    - name: run-coverage
+      image: gcr.io/tekton-releases/dogfooding/coverage:latest
```
afrittoli, line 58: NIT: perhaps we should use a sha instead of latest

tekton/ci/shared/go_coverage.yaml
chitrangpatel: sounds good! I will move it there.

tekton/ci/shared/go_coverage.yaml
chitrangpatel: I don't think so. Just an oversight. I will move it to 0.3

chitrangpatel: done

tekton/ci/shared/go_coverage.yaml
chitrangpatel: I just wasn't aware of it. I updated it to use git-batch-merge instead.

tekton/ci/shared/go_coverage.yaml
chitrangpatel: done

tekton/ci/shared/go_coverage.yaml
chitrangpatel: I was keeping it consistent with what we used in the prow job: https://github.com/tektoncd/plumbing/blob/fa0f5c274777261077164e4cc612195d883b480a/prow/config.yaml#L1223

I'm happy to change it. But for the purposes of testing, we should keep it the same so that we can compare then coverages apples-to-apples. WHYT?

dibyom:
Aside from using a fixed SHA vs latest, is there anything else we need to do for this PR?

chitrangpatel:
> Aside from using a fixed SHA vs latest, is there anything else we need to do for this PR?

I don't think so. I've addressed all other comments.

afrittoli (commented):
Thanks @chitrangpatel 
/lgtm

This migrates the presubmit job. 
Do you plan on migrating the post-submit job as well? 
If I understand correctly, the post-submit job updates the baseline against which the coverage deltas are calculated.

PR #1250, opened 2022-10-25
Skip/fail commenting if the job status does not match the report input

abayer:
# Changes

Yet another attempt to get the PR commenter to stop posting pointless comments - here, we take advantage of the fact that we've already set the commit status for the job (via `pr-status-updater`) before we handle the possible comment. That means that if we get the status from the API and it doesn't match the input, we just skip.

NOTE: Now that I've actually implemented this, I'm not sure if there's a point - whatever's causing the duplicate event triggers is calling `pr-status-updater` with the same input that's passed to `pr-commenter`, so in practice, these should always match. Hrm.

/kind misc

# Submitter Checklist

review them:

for more details._


afrittoli:
@abayer Do you still want to do this?

abayer:
Nah, this isn’t necessary.

PR #1279, opened 2022-11-11
[Upgrade test] Update config for upgrade test

JeromeJu:
# Changes
This commit migrates upgrade test from relying on GKE only to be 
using kind cluster setup. And this follows the guide in docs/kind-e2e.md.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1846,6 +1846,9 @@ periodics:
   name: ci-tekton-pipeline-upgrade-tests
   labels:
     preset-presubmit-sh: "true"
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
+    pipeline-kind-e2e: "true"
   agent: kubernetes
   extra_refs:
   - org: tektoncd
@@ -1853,22 +1856,49 @@ periodics:
     base_ref: main
     path_alias: github.com/tektoncd/pipeline
   spec:
+    affinity:
+      podAntiAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+                - key: pipeline-kind-e2e
+                  operator: Exists
+            topologyKey: kubernetes.io/hostname
+    nodeSelector:
+      cloud.google.com/gke-ephemeral-storage-local-ssd: "true"
+      cloud.google.com/gke-nodepool: n2-standard-4-kind
+    tolerations:
+      - key: kind-only
+        operator: Equal
+        value: "true"
+        effect: NoSchedule
     containers:
     - image: gcr.io/tekton-releases/dogfooding/test-runner@sha256:01bace96516d22d70c092ff9c01f9ffe7af806c7475157dcaa7cdb45d84bd68b # golang 1.18.7
       imagePullPolicy: Always
       command:
         - /usr/local/bin/entrypoint.sh
       args:
-      - "--scenario=kubernetes_execute_bazel"
-      - "--clean"
-      - "--job=$(JOB_NAME)"
-      - "--repo=github.com/tektoncd/pipeline=main"
-      - "--root=/go/src"
       - "--service-account=/etc/test-account/service-account.json"
-      - "--upload=gs://tekton-prow/pr-logs"
       - "--" # end bootstrap args, scenario args below
       - "--" # end kubernetes_execute_bazel flags (consider following flags as text)
+      - "/usr/local/bin/kind-e2e"
+      - "--k8s-version"
+      - "v1.23.x"
+      - "--nodes"
+      - "3"
+      - "--e2e-script"
       - "./test/e2e-tests-upgrade.sh"
+      - "--e2e-env"
+      - "./test/e2e-tests-kind-prow.env"      
+      securityContext:
+        privileged: true
+      resources:
+        requests:
+          cpu: 3500m
+          memory: 4Gi
+        limits:
+          cpu: 3500m
+          memory: 8Gi 
       env:
       - name: SKIP_INITIALIZE
         value: "true"


prow/config.yaml
```diff
@@ -1846,29 +1846,60 @@ periodics:
...
+      - "/usr/local/bin/kind-e2e"
+      - "--k8s-version"
+      - "v1.23.x"
+      - "--nodes"
+      - "3"
+      - "--e2e-script"
+      - "./test/e2e-tests.sh"
+      - "--e2e-env"
+      - "./test/e2e-tests-kind-prow.env"      
       - "./test/e2e-tests-upgrade.sh"
```
JeromeJu, line 1893: Wondering if there is a easy way to test for validating the args here with `docker run`

dibyom (approved):

prow/config.yaml
```diff
@@ -1846,29 +1846,60 @@ periodics:
...
+    pipeline-kind-e2e: "true"
   agent: kubernetes
   extra_refs:
   - org: tektoncd
     repo: pipeline
     base_ref: main
     path_alias: github.com/tektoncd/pipeline
   spec:
+    affinity:
+      podAntiAffinity:
```
lbernick, line 1860: It seems like this is saying it cannot run on the same node as pipeline e2e tests, why is this? can you add some comments here? Same goes for the combinations of nodeSelector and tolerations.

prow/config.yaml
```diff
@@ -1846,29 +1846,60 @@ periodics:
...
-      - "--job=$(JOB_NAME)"
-      - "--repo=github.com/tektoncd/pipeline=main"
-      - "--root=/go/src"
       - "--service-account=/etc/test-account/service-account.json"
-      - "--upload=gs://tekton-prow/pr-logs"
       - "--" # end bootstrap args, scenario args below
       - "--" # end kubernetes_execute_bazel flags (consider following flags as text)
+      - "/usr/local/bin/kind-e2e"
+      - "--k8s-version"
+      - "v1.23.x"
```
lbernick, line 1886: Do we have a process for changing what k8s version our tests are run on? How will we ensure that people remember to update this value?

prow/config.yaml
```diff
@@ -1846,29 +1846,60 @@ periodics:
...
+      - "--nodes"
+      - "3"
+      - "--e2e-script"
+      - "./test/e2e-tests.sh"
+      - "--e2e-env"
+      - "./test/e2e-tests-kind-prow.env"
```
lbernick, line 1892: Does this run e2e tests in addition to upgrade tests? (I think upgrade tests include e2e tests)

prow/config.yaml
lbernick: Discussed offline but I think it's worth exploring if this could be a taskrun triggered by a cron job as it'll be much easier to test before merging.

prow/config.yaml
```diff
@@ -1846,29 +1846,60 @@ periodics:
...
+      - "v1.23.x"
+      - "--nodes"
+      - "3"
+      - "--e2e-script"
+      - "./test/e2e-tests.sh"
+      - "--e2e-env"
+      - "./test/e2e-tests-kind-prow.env"      
       - "./test/e2e-tests-upgrade.sh"
+      securityContext:
+        privileged: true
```
lbernick, line 1895: double checking that this is needed for kind?

dibyom:
@lbernick I think most of this is just following the instructions in https://github.com/tektoncd/plumbing/blob/main/docs/kind-e2e.md

prow/config.yaml
dibyom: IMO, we should split off getting upgrade tests to work from migrating them to tekton

dibyom: Also, just moving to kind makes it easy enough to run these scripts against a local kind cluster

lbernick:
>  I think most of this is just following the instructions in https://github.com/tektoncd/plumbing/blob/main/docs/kind-e2e.md

Ah thanks for helping me read :) I don't see pod affinity in there so I'm still curious why that's needed, and I'm still confused why we'd need the e2e test script in addition to the upgrade test script in the args. The reason I suggested moving this to Tekton is to make it easier to test. I'm ok with merging this as a prow job first but I think we want to make sure it works (to the extent that we can do so) before merging it.

dibyom:
>  why we'd need the e2e test script in addition to the upgrade test script in the args. 

Oh, you are right, we don't - the e2e-upgrade script should be the test script - @JeromeJu we should fix that

dibyom:
>  I don't see pod affinity in there so I'm still curious why that's needed,

I think the anti-affinity got added because two kind e2e tests running on the same node would impact each other: https://github.com/tektoncd/plumbing/pull/1135/ - we should definitely add this to the docs!

prow/config.yaml
dibyom: we do unfortunately

JeromeJu:
> > why we'd need the e2e test script in addition to the upgrade test script in the args.
> 
> Oh, you are right, we don't - the e2e-upgrade script should be the test script - @JeromeJu we should fix that

Thanks, updated!

prow/config.yaml
```diff
@@ -1996,3 +2026,5 @@ periodics:
       - --template
       - --ceiling=10
       - --confirm
+
+docker run -it test-runner-1
```
dibyom, line 2030: This line looks like it was added unintentionally?

prow/config.yaml
JeromeJu: Thanks to the pointers from @abayer , this is tested with `docker run` with privileged mode and docker socket passed in and then the `kind-e2e.sh` script could be started with the respective inputs as in the config.yaml. And thus this PR could be regarded as tested before merging.

prow/config.yaml
JeromeJu: Thanks to @dibyom , this is added at https://github.com/tektoncd/plumbing/issues/1281

vdemeester (approved):

prow/config.yaml
JeromeJu: Thanks Lee, opened #1283 to keep track of this.

PR #1308, opened 2022-12-22
Upgraded golangci-lint to 1.50.1 (latest).

bendory:
# Changes

Upgraded golangci-lint to 1.50.1 (latest).

This will make some new linters (like `dupword`) available; see https://github.com/tektoncd/pipeline/pull/5918.

Context: https://github.com/tektoncd/pipeline/issues/5899

/kind cleanup

# Submitter Checklist

- [N/A] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

PR #1309, opened 2022-12-28
Upgraded `test-runner` image to latest.

bendory:
# Changes

This change will update our plumbing to include
https://github.com/tektoncd/plumbing/pull/1308
as well as any other changes since the last upgrade.

Context: https://github.com/tektoncd/pipeline/pull/5918#issuecomment-1366766687
Issue: https://github.com/tektoncd/pipeline/issues/5899

/kind cleanup

# Submitter Checklist

review them:

- [N/A] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -395,7 +395,7 @@ presubmits:
       trigger: "(?m)^/test (all|pull-tekton-catlin-build-tests),?(\\s+|$)"
       spec:
         containers:
-        - image: gcr.io/tekton-releases/dogfooding/test-runner@sha256:01bace96516d22d70c092ff9c01f9ffe7af806c7475157dcaa7cdb45d84bd68b # golang 1.18.7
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -1165,7 +1165,7 @@ presubmits:
     trigger: "(?m)^/test (all|pull-tekton-pipeline-build-tests),?(\\s+|$)"
     spec:
       containers:
-      - image: gcr.io/tekton-releases/dogfooding/test-runner@sha256:01bace96516d22d70c092ff9c01f9ffe7af806c7475157dcaa7cdb45d84bd68b # golang 1.18.7
```
abayer, line 1168: It would probably be worth replacing this SHA in all the jobs.

abayer (changes requested):

prow/config.yaml
bendory: Doh! I naively thought this would be defined in one place and used everywhere.

PR #1310, opened 2023-01-03
Bump go to 1.18.9 in "all" images.

vdemeester:
# Changes

Going from 1.18.7 to the latest bugfix of go 1.18.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/images/coverage/Dockerfile
+++ b/tekton/images/coverage/Dockerfile
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.18.7 as buildcoverage
+FROM golang:1.18.9 as buildcoverage
 RUN git clone https://github.com/knative/test-infra /go/src/knative.dev/test-infra
 RUN git -C /go/src/knative.dev/test-infra checkout ba4c2c3e061a59ac4b167da84924e9fa55475ad9 # Last commit before removal of tools/coverage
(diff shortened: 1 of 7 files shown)


vdemeester:
@afrittoli somehow, something is wrong with `plumbing-image-build` pipeline 🤔

vdemeester:
@afrittoli @abayer it's finally green 🎉 (thanks for fixing the CI)

afrittoli:
> @afrittoli @abayer it's finally green 🎉 (thanks for fixing the CI)

The image build is now passing, however it runs only for the first image in the list.
This is because the taskloop controller was not maintained anymore, because we have matrix, but matrix is not functionally equivalent, it does not take a variable length array as input (like the results of the previous task), so the only options were:
- build all images all the time
- build the first image in the list

If we had one image changed per PR this would actually work, but on this PR we're only testing the first image.
I guess something like the old mario service could be used as a workaround for this kind of PRs, but I don't think it's in an healthy state

afrittoli:
> /mario build tekton/images/kubectl kubectl:mario

Mario was migrated to use the pipeline-in-a-pod custom task which only understands `Runs` and the pipeline v0.44 now produces `CustomRun` so no chance for it to run as it is.

vdemeester:
@afrittoli do you want me to split this PR into several ones ? (one per Dockerfile?)

vdemeester:
Alright, I'll probably close and propose a PR per bump, and most likely will bump to go 1.19 as well.

PR #1317, opened 2023-01-13
Fix perms for custom tasks, revert those custom tasks back to v1alpha1.Runs for now

abayer:
# Changes

This is two commits on purpose to make it easier to revert the `CustomRun`->`Run` change later.

First, I added `customruns` to the permissions for the `pr-commenter` and `pr-status-updater` service accounts, which was kinda needed if they're trying to watch `CustomRun`s. =)

But once I fixed that in-cluster manually, I realized that the custom tasks invoked by various triggered `PipelineRun`s for these particular custom tasks...were being generated as `Run`s. That's because the `PipelineRun` reconciler only recently added a feature flag to enable creating `CustomRun`s instead, and that's not switching to the default until v0.44, if I remember correctly. So for now, we need to go back to `Run`s here.

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/ci/custom-tasks/pr-commenter/README.md
+++ b/tekton/ci/custom-tasks/pr-commenter/README.md
@@ -18,8 +18,8 @@ deployment for that to be reflected in the comment.
 ## Example `Run`
 
 ```yaml
-apiVersion: tekton.dev/v1beta1
-kind: CustomRun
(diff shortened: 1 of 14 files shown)


afrittoli (commented):
Thank you!
/approve

PR #1323, opened 2023-01-26
Verified Catalog E2E Test Common Scripts

QuanZhang-William:
Part of [Verified Catalog Migration][#3]. This commit adds a new script containing the common helper script code for [Verified Catalogs][Verified Catalog].

The content is based on: https://github.com/tektoncd/catalog/blob/main/test/e2e-common.sh, with improvements/refactoring for the new catalog git-based versioning and decentralized-repo model contracts.

The script supports both on PR merge integration test and nightly integration tests, specified by `TEST_RUN_NIGHTLY_TESTS` env variable.

In `nightly test`, the script runs test cases in all release of the catalog. 
In `on-PR-merge test`, the script runs test cases in the branch head.

An e2e use example can found in: https://github.com/QuanZhang-William/golang/tree/test-pipelines.

[#3]: https://github.com/tektoncd-catalog/.github/issues/3
[Verified Catalog]: https://github.com/tektoncd-catalog

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/scripts/verified-catalog-e2e-common.sh
+++ b/scripts/verified-catalog-e2e-common.sh
@@ -1,0 +1,301 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Helper functions for E2E tests.
+source $(dirname ${BASH_SOURCE})/e2e-tests.sh
+
+# Define a custom kubectl path if you like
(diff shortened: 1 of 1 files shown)


scripts/verified-catalog-e2e-common.sh
```diff
@@ -0,0 +1,301 @@
...
+            cnt=0
+            task_to_tests=""
+            continue
+        fi
+        cnt=$((cnt+1))
+    done
+
+    # in case if there are some remaining tasks
+    if [[ -n ${task_to_tests} ]];then
+        test_task_creation "${task_to_tests}"
```
vinamra28, line 253: I guess we don't need this look now. This was required when no of tasks within catalog were increasing at a good amount of rate but now since we are moving to decentralised catalogs, we don't need this

QuanZhang-William: In nightly test we will test all the versions/releases of the catalog, which could scale to >8 tests (in the future), so I was thinking to keep this so that we can have some safety limit in the test when we have multiple versions in the future? 🤔

vinamra28: but in that case, should we not just test the latest 3 versions of the applications?

QuanZhang-William: In the current design of TEP-0079 and TEP-0115, I don't think we called out that we will only support 3 latest versions of the catalog, so I think we should test all of them. But we can definitely further discuss this option later.

vdemeester (approved):

PR #1393, opened 2023-04-14
increase TaskRunTimeout for s390x dashboard tasks

Basavaraju-G:
e2e-dashboard tests are running in docker-in-docker. its taking long time to complete. so increase task and pipeline run timeout for s390x tekton dashboard jobs.

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/tekton/resources/nightly-tests/dashboard-deploy-test-s390x-template.yaml
+++ b/tekton/resources/nightly-tests/dashboard-deploy-test-s390x-template.yaml
@@ -32,7 +32,6 @@ spec:
       namespace: $(tt.params.namespace)
     spec:
       timeout:
-        pipeline: "3h4m0s"
         tasks: "2h0m0s"
       workspaces:
       # this workspace will be used to share info between tasks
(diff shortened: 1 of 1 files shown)


AlanGreene:
As mentioned on Slack, instead of increasing the pipeline timeout, how about removing the 1 hour timeout on the tasks and letting them take as long as needed within the current pipeline timeout?

Separately we'll also need to improve the performance of these runs as they should not be taking 2 hours to complete.

For reference, here are the resource requests and limits we use when running these tests on prow: https://github.com/tektoncd/plumbing/blob/47ba204a8771bb78613a6066dbf70866bebcc089/prow/config.yaml#L862-L868

AlanGreene:
Confirmed on dogfooding that the change in https://github.com/tektoncd/plumbing/pull/1394 (running Z and P tests at different times, and switching storage driver) consistently gets us down from 1 hour for the e2e task to about 30 minutes per attempt so we should be ok with the current 2 hour timeout.

Let's see what happens tomorrow morning when these run again and decide whether we need to adjust the timeouts. 🤞

PR #1521, opened 2023-07-27
Update test-runner image in prow config

chuangw6:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/6992

Using the latest built image
`gcr.io/tekton-releases/dogfooding/test-runner@sha256:11dc05cf6e81786ce5935e43aad29a84d6b662804cb57f740c9fb3e1e73ff5f7`

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1287,7 +1287,7 @@ presubmits:
           value: "true"
           effect: NoSchedule
       containers:
-        - image: gcr.io/tekton-releases/dogfooding/test-runner@sha256:5e5c4b04b7e7459f306eb0b8424495652b034b4b06a9e9ea7b0fe22282fd11af # golang 1.19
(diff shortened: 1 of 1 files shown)


chuangw6:
cc @vdemeester @lbernick

lbernick (commented):
/lgtm
Thanks Chuang!

lbernick:
> Not sure if we should update all test-runner image to the latest in the prow config. If it's fine to change, please let me know.

Let's just fix the problem with existing CI, and we can upgrade other projects' prow config separately.

vdemeester:
/approve
/lgtm

I agree with @lbernick. I think ideally we would have something like dependabot that updates the images, runs some "quick test" with components and gets automatically merged if it's green... But ideally, for this, we would probably want/need to have the CI setup directly in each components instead of here… (because here, we might need to update the image for pipeline to make CI green, but updating would break cli or another component.. which is definitely not an ideal situation)

PR #1970, opened 2024-05-14
Fix peribolos secret in the tekton-cd el

afrittoli:
# Changes

When peribolos is run via cron, it uses a different template, so we must update the secret there too.

# Submitter Checklist

review them:

for more details._

/kind misc


vinamra28 (commented):
/lgtm

PR #2066, opened 2024-07-04
tekton/images: fetch `ko` from upstream image.

vdemeester:
# Changes
That way, a tool like dependabot or renovate will automatically update
the ko version for us.

/kind misc

# Submitter Checklist

review them:

for more details._


vdemeester:
cc @afrittoli 
I think we can even remove the `ko` image as it could be replaced where we use it (in the `publish.yaml` during release). For `ko-gcloud` it might be a bit trickier, it seems to be used in `install-tekton-release` but not sure why/how as we don't use `ko` at all in there it seems.

PR #2330, opened 2024-12-06
Reusable actions to restest GitHub actions

afrittoli:
# Changes

/kind feature

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/.github/actions/_chatops_retest.yml
+++ b/.github/actions/_chatops_retest.yml
@@ -1,0 +1,93 @@
+# The _chatops_retest workflow reruns failed GHA for a PR
+#
+# This workflow is triggered by leaving a "/retest" comment on
+# a pull request. If the required preconditions are met, it will
+# rerun failed GitHub actions checks on that PR
+#
+# Condition for the "/retest" command are:
+# - either the issuer is a maintainer
+# - or the issuer is the owner the PR
+#
+# To use this in a repo:
+#
+# name: Rerun Failed Actions
+# on:
+#   repository_dispatch:
+#     types: [retest-command]
+#
+# jobs:
+#   retest:
+#     name: Rerun Failed Actions
+#     uses: tektoncd/plumbing/.github/actions/_chatops_retest.yml@main
+
+name: Rerun Failed Actions
+on:
+  workflow_call:
+
+jobs:
+  retest:
+    name: Rerun Failed Actions
+    runs-on: ubuntu-latest
+    steps:
+    - name: Show Environment Variables
+      run: env
+    - name: Show Github Object
+      run: |
+        cat <<'EOF'
+        ${{ toJson(github) }}
+        EOF
+    - name: Show Github Event Path Json
+      run: 'cat $GITHUB_EVENT_PATH || true'
+    - name: Rerun Failed Actions
+      id: rerunFailed
+      run: |
+        echo '::group:: Get the PR commit sha'
+        # Get the sha of the HEAD commit in the PR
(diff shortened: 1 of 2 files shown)


.github/actions/_slash.yml
```diff
@@ -0,0 +1,51 @@
...
+# Slash commands must be placed at the very beginning of the
+# first line of a comment. More details are available in the
+# action docs: https://github.com/peter-evans/slash-command-dispatch/tree/main?tab=readme-ov-file#how-comments-are-parsed-for-slash-commands
+#
+# The workflow looks for and dispatches to another workflow
+# named <command>-command which must exist in the repository.
+#
+# Supported commands:
+# - /land: invokes the land-command workflow, to land (merge) PRs
+#          stacked through ghstack
```
AlanGreene, line 16: This should be replaced with our `retest` command, or just reference the other action file instead since that already includes documentation.

.github/actions/_chatops_retest.yml
```diff
@@ -0,0 +1,92 @@
...
+        done
+        echo '::endgroup::'
+      env:
+        GITHUB_TOKEN: ${{ secrets.CHATOPS_TOKEN }}
+        GITHUB_REPO: ${{ github.event.client_payload.github.payload.repository.full_name }}
+        GITHUB_PULL_URL: ${{ github.event.client_payload.github.payload.issue.pull_request.url }}
+
+    - name: Create comment
+      if: ${{ failure() && steps.landStack.outcome == 'failure' }}
+      uses: peter-evans/create-or-update-comment@v4
```
AlanGreene, line 75: we should probably pin actions by sha and let dependabot take care of updating them as needed

AlanGreene (approved):
Overall this looks good, just a few minor comments below.

Would it be worth also adding [workflow templates](https://docs.github.com/en/actions/sharing-automations/creating-workflow-templates-for-your-organization) for these in tektoncd/.github to make them easier to adopt in the other projects?

vdemeester (approved):

.github/actions/_chatops_retest.yml
```diff
@@ -0,0 +1,92 @@
...
+                --failed || true
+        done
+        echo '::endgroup::'
+      env:
+        GITHUB_TOKEN: ${{ secrets.CHATOPS_TOKEN }}
+        GITHUB_REPO: ${{ github.event.client_payload.github.payload.repository.full_name }}
+        GITHUB_PULL_URL: ${{ github.event.client_payload.github.payload.issue.pull_request.url }}
+
+    - name: Create comment
+      if: ${{ failure() && steps.landStack.outcome == 'failure' }}
```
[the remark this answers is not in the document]
afrittoli: Good catch, fixed now

afrittoli:
> Overall this looks good, just a few minor comments below.
> 
> Would it be worth also adding [workflow templates](https://docs.github.com/en/actions/sharing-automations/creating-workflow-templates-for-your-organization) for these in tektoncd/.github to make them easier to adopt in the other projects?

In this case I prefer a reusable actions to a workflow template. 

This action can be used directly by other repos and can be pinned to a specific sha there to ensure consistency in CI results. Templates serve as a starting point to creating new workflows but diverge over time from the template. 

For a workflow to benefit from changes made here, it has to:
* with reusable action: update the sha
* with templates: look for changes in the template and apply them manually in the local workflow

Template give downstream users flexibility to do local changes, but given that the reusability is only meant within the Tekton community, I don't think this would be a big advantage.

PR #2689, opened 2025-07-16
.github/workflows: add a "build" image workflow

vdemeester:
# Changes

/kind misc

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -1,0 +1,83 @@
+name: build-images
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 0 * * *"
+    
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull-request.number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  list:
+    runs-on: ubuntu-latest
+    outputs:
+      images: ${{ steps.output-images.outputs.images }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: list images
+        id: output-images
+        run: |
+          echo "images=$(find tekton/images/* -maxdepth 1 -type d -printf '%f\n' | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
+
+  build:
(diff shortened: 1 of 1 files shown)


.github/workflows/images.yaml
```diff
@@ -0,0 +1,85 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@cf79a64fed8a943fb1073260883d08fe0dfb4e56 # v45.0.7
+        with:
+          files: |
+            tekton/images/${{ matrix.image }}/**
+      - name: changed files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: |
+          echo ${{ steps.changed-files.outputs.any_changed }}
+          echo "List all the files that have changed: ${{ steps.changed-files-specific.outputs.all_changed_files }}"
```
afrittoli, line 64: I wonder if it would be possible to move this to the list.
I'm not 100% sure how the matrix is implemented by GHA, but I assume it will use one runner per branch in the matrix, so it may be allocating 10 or so runners for every PR that may not even include any change to any image

.github/workflows/images.yaml
```diff
@@ -0,0 +1,85 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        if: steps.changed-files.outputs.any_changed == 'true' || github.event_name != 'pull_request'
+        id: meta
+        uses: docker/metadata-action@902fa8ec7d6ecbf8d84d538b9b233a880e428804 # v5.7.0
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository }}/${{ matrix.image }}
+      - name: Build and push Docker image
+        if: steps.changed-files.outputs.any_changed == 'true' || github.event_name != 'pull_request'
```
afrittoli, line 79: Doesn't this mean that on merge, we'll run the build even if there was no change?

afrittoli (commented):
Thanks for this!
Looks good, just a couple of comments

.github/workflows/images.yaml
```diff
@@ -0,0 +1,85 @@
...
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v5
+        with:
+          context: tekton/images/${{ matrix.image }}
+          push: ${{ github.event_name != 'pull_request' }}
```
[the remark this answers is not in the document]
vdemeester: yes

.github/workflows/images.yaml
vdemeester: After discussing this offline, let's do this if we run into issues.

PR #3017, opened 2025-12-17
fix: Improve cherry-pick workflow error handling

vdemeester:
# Changes

This PR improves error handling in the cherry-pick workflow to prevent false success reports and enable better debugging.

**What was broken:**
The workflow would report "✅ Cherry-pick completed successfully!" even when git push or PR creation failed. This happened because:
- `git push` failures weren't checked (e.g., token missing `workflow` scope)
- `gh pr create` failures weren't checked (e.g., "Head sha can't be blank")

**What this fixes:**
- Wrap `git push` in conditional check that exits with error if push fails
- Capture `gh pr create` output and check exit code, failing appropriately
- Display the created PR URL in success comment for immediate access

**Example of the old broken behavior:**
```
Pushing cherry-pick branch...
To https://github.com/tektoncd/operator
 ! [remote rejected] cherry-pick-3106-to-release-v0.78.x (refusing to allow a Personal Access Token to create or update workflow without `workflow` scope)
Creating pull request...
pull request create failed: GraphQL: Head sha can't be blank
✅ Cherry-pick completed successfully!
```

With this fix, the workflow will properly detect these failures and report them in the PR comment.

/kind bug

# Submitter Checklist

- [x] Includes docs (if user facing) - N/A, internal workflow improvement
- [x] Commit messages follow commit message best practices

### DIFF
--- a/.github/workflows/_cherry-pick-command.yaml
+++ b/.github/workflows/_cherry-pick-command.yaml
@@ -112,10 +112,11 @@ jobs:
           TARGET_BRANCH: ${{ matrix.branch }}
           PR_NUMBER: ${{ github.event.client_payload.pull_request.number }}
         run: |
-          set +e  # Don't exit on error, we want to capture it
+          set -e  # Exit on error to catch unexpected failures
 
           # Capture all output
           OUTPUT_FILE=$(mktemp)
+          ERROR=""
 
           {
             echo "🤖 Starting cherry-pick process..."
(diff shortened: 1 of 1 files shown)


.github/workflows/_cherry-pick-command.yaml
afrittoli: I think a safer approach would be to use `set -e` instead, and capture the return code where needed, e.g.

```
command || ERROR="true"
if [[ "ERROR" == "true' ]]; then
    <error handling>
...
```

This way, in places where we have unexpected failures the script will still report failure instead of reporting success.

Issue #3038, opened 2025-12-30
Running tekton_in_kind.sh hangs and does not return control to Bash on macOS


PetronellaPC:
# Expected Behavior
After the tekton_in_kind.sh script finishes running (whether it succeeds or fails), control should return to the Bash shell so the user can continue using the terminal normally.

# Actual Behavior
When running the https://github.com/tektoncd/plumbing/blob/main/hack/tekton_in_kind.sh script on macOS in a bash shell, the script reaches completion but does not return control to the bash shell.

The terminal appears to be stuck after the script finishes.
CTRL+C does not interrupt the process or return control.
The only way to recover is to close the terminal window and open a new one.

This behavior occurs regardless of whether the script execution succeeds or fails.

# Steps to Reproduce the Problem

Open Terminal on macOS.

Run the hack/tekton_in_kind.sh script from the tektoncd/plumbing repository using Bash:

`. tekton_in_kind.sh `

Wait for the script to complete (successfully or unsuccessfully) and observe that the terminal does not return to the Bash prompt.

# Additional Info

Kubernetes version:
Output of kubectl version:
Not relevant

Tekton Pipeline version:
Output of tkn version or kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'
Not relevant

Operating System:
macOS 26.2 (25C56)

Shell:
Bash

PetronellaPC:
Additional context (possible contributing factor)

While investigating this issue, I noticed a difference in behavior depending on how the script is invoked.

When the script is executed with `./`:

`./hack/tekton_in_kind.sh`

the shell returns control as expected after completion.

However, when the script is run via dot-sourcing:

`. hack/tekton_in_kind.sh`

the terminal becomes unusable after the script completes (no prompt is returned, and CTRL+C does not work), requiring the terminal to be closed.

Given that the script performs process management, sets shell options, and may define traps or spawn background processes, this behavior is understandable when sourced, but it may not be immediately obvious to users that dot-sourcing is unsupported.

Suggested improvement:

To make the behavior clearer and avoid accidental misuse, it might be helpful to add a small guard at the top of the script to detect when it is being sourced and exit early with a clear message, for example:

```
if [[ "${BASH_SOURCE[0]}" != "$0" ]]; then
  echo "This script is not intended to be sourced. Please run it as ./tekton_in_kind.sh"
  return 1
fi
```

[...]

PR #3039, opened 2026-01-02
Prevent sourcing of tekton_in_kind.sh Issue: 3038

PetronellaPC:
The tekton_in_kind.sh script can currently be sourced without error, which may lead to unexpected behavior such as early returns, partially executed logic, or confusing failures for users who expect it to run as a standalone entrypoint.

Adding an explicit guard to detect sourcing and fail fast provides clearer feedback and avoids subtle misuse. Alternatives such as documenting correct usage or relying on downstream failures were considered, but enforcing correct invocation in the script itself is more robust and user-friendly. This change has no impact on valid usage and leaves room for future refactoring if sourcing support is ever intentionally added. Addressing issue https://github.com/tektoncd/plumbing/issues/3038

# Changes

This PR prevents tekton_in_kind.sh from being sourced by mistake by
explicitly detecting when the script is sourced and failing fast with a
clear error message.

Sourcing this script can lead to confusing behavior (such as unexpected
return statements or partially executed logic), since it is designed
to be run as a standalone executable. By enforcing correct invocation,
the script becomes safer and more user-friendly, addressing the problem
described in https://github.com/tektoncd/plumbing/issues/3038

/kind bug

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/hack/tekton_in_kind.sh
+++ b/hack/tekton_in_kind.sh
@@ -16,6 +16,13 @@ declare TEKTON_PIPELINE_VERSION TEKTON_TRIGGERS_VERSION TEKTON_DASHBOARD_VERSION
 # - If a kind cluster named "tekton" already exists this will fail
 # - Local access to the dashboard requires port 9097 to be locally available
 
+# if script is ran with source, i.e. `. tekton_in_kind.sh` - a warning is displayed 
+# and suggested to run it in it's own shell process. 
+if [[ "${BASH_SOURCE[0]}" != "$0" ]]; then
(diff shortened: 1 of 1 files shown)


twoGiants (approved):
Hi @PetronellaPC! :wave: :smile_cat: 

Thank you for you contribution and welcome to Tekton!

The change looks good to me :+1: 

/approve 
/lgtm

twoGiants:
/cc @afrittoli @AlanGreene 

It's a quick one! Would be good to get a **lgtm**. Thanks :smile_cat: :+1:

vdemeester (approved):
/lgtm

PR #3221, opened 2026-03-15
feat: add static base image build for Tekton Go binaries

waveywaves:
# Changes

Add apko-based build for a multi-arch static base image supporting amd64, arm64, s390x, and ppc64le.

**Why:** All Tekton repos currently depend on `cgr.dev/chainguard/static` which is pinned to an EOL Alpine 3.18 digest from Nov 2023. The Chainguard free tier dropped s390x/ppc64le support, so bumping the digest breaks release builds. Building our own image gives us control over architecture coverage and update cadence.

**What this adds:**
- `images/static-base/apko.yaml` — declarative image definition, builds a ~300KB static base per arch with CA certs, timezone data, and nonroot user
- `.github/workflows/build-static-base-image.yaml` — weekly CI to rebuild and publish to `ghcr.io/tektoncd/plumbing/static-base`
- `images/static-base/README.md` — documents contents, build instructions, and consumers

**Consumers:** tektoncd/pipeline, tektoncd/triggers, tektoncd/chains, tektoncd/results — all can reference `ghcr.io/tektoncd/plumbing/static-base` as their `defaultBaseImage` in `.ko.yaml`.

Related: tektoncd/pipeline#9557

/kind feature

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

PR #3434, opened 2026-06-09
feat: add apko-based static base image for Tekton Go binaries

vdemeester:
# Changes

Carries #3221 (by @waveywaves) and addresses the review feedback.

Adds an apko-based multi-arch static base image (amd64, arm64, s390x, ppc64le) for all Tekton Go binaries, built and published as part of the existing `ci` workflow.

**Why:** Tekton repos currently depend on `cgr.dev/chainguard/static`, pinned to an EOL Alpine 3.18 digest from Nov 2023. The Chainguard free tier dropped s390x/ppc64le support, so bumping the digest breaks release builds. Building our own image gives us control over architecture coverage and update cadence. See tektoncd/pipeline#9557.

**What this adds:**
- `tekton/images/static-base/apko.yaml` — declarative image definition (~300KB static base per arch with CA certs, timezone data, nonroot user)
- `tekton/images/static-base/README.md` — contents, build instructions, consumers
- `ci.yaml` `build-image` job now builds apko images alongside Dockerfile images

## Changes vs the original #3221 (review feedback)

- **Integrated into `ci.yaml` instead of a standalone workflow** (per @vdemeester). The `build-image` job detects an `apko.yaml` (vs a `Dockerfile`) and builds with apko: it validates on `pull_request` and publishes to `ghcr.io/tektoncd/plumbing/static-base` on push to `main` and on the daily schedule. This keeps it gated by the `CI summary` check and keeps all images under `tekton/images/`. The dedicated `build-static-base-image.yaml` workflow is removed.
- **SHA-pinned** the `setup-apko` action to `v1.6.22` (per @anithapriyanatarajan).
- **Completed the apko keyring** with one signing key per target architecture (per @vdemeester re: keyring validation). Alpine signs each arch's `APKINDEX` with a different key — with only the x86_64 key, `ppc64le` (and, in a clean CI environment, `arm64`/`s390x`) signature verification fails. Verified a full 4-arch `apko build` succeeds with all keys present.

**Consumers:** tektoncd/pipeline, tektoncd/triggers, tektoncd/chains, tektoncd/results — all can reference `ghcr.io/tektoncd/plumbing/static-base` as their `defaultBaseImage` in `.ko.yaml`.

Related: tektoncd/pipeline#9557
Carries: #3221

/kind feature

# Submitter Checklist


tekton/images/static-base/README.md
```diff
@@ -0,0 +1,66 @@
...
+| `/etc/passwd`, `/etc/group` | nonroot user (UID 65532) |
+| `/etc/nsswitch.conf` | DNS resolution |
+
+## Build locally
+
+Run from the repository root so the vendored keyring paths in `apko.yaml`
+resolve correctly:
+
+```bash
+# Install apko: go install chainguard.dev/apko@latest
```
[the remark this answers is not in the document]
vdemeester: Good catch — pinned the local-build instructions to `apko@v1.2.16` to match `APKO_VERSION` in ci.yaml in df05353.

waveywaves (commented):
Updated after re-checking the original context in tektoncd/pipeline#9557 and the current Chainguard replacement image.

No blocker from my side; I approved separately. The current state I think the PR should reflect is:

- This PR is intentionally **Alpine-based**: apko + Alpine APK repositories + vendored Alpine signing keys.
- No architecture implementation change is needed; the PR builds the required Tekton arches: `amd64`, `arm64`, `s390x`, and `ppc64le`.
- `ghcr.io/wolfi-dev/static:alpine` exists and currently supports those arches, so the rationale should not be “there is no updated multi-arch Alpine static image”.
- The better rationale is that Tekton wants to **own/build/control** the static base image in `tektoncd/plumbing`: exact package set, signing keys, rebuild cadence, CI gating, SBOM generation, and published GHCR image/tag/digest flow.

Suggested PR-description wording:

```md
While `ghcr.io/wolfi-dev/static:alpine` is available as a maintained Alpine-based replacement for the old `cgr.dev/chainguard/static` image and currently supports the architectures Tekton needs, this PR intentionally builds the static base image in `tektoncd/plumbing` so Tekton owns the package set, architecture matrix, signing keys, rebuild cadence, CI gating, SBOM generation, and published image location.

This keeps Tekton release artifacts from depending on external static-base image publication policy or availability, while still using the Alpine/apko approach.
```

I left the concrete notes inline: add `alpine-release` if we want Alpine identity metadata, and optional follow-ups for OCI metadata/tag consistency.

tekton/images/static-base/apko.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - tekton/images/static-base/keys/alpine-devel@lists.alpinelinux.org-6165ee59.rsa.pub  # x86_64
+    - tekton/images/static-base/keys/alpine-devel@lists.alpinelinux.org-616ae350.rsa.pub  # aarch64
+    - tekton/images/static-base/keys/alpine-devel@lists.alpinelinux.org-616ac3bc.rsa.pub  # s390x
+    - tekton/images/static-base/keys/alpine-devel@lists.alpinelinux.org-58cbb476.rsa.pub  # ppc64le
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/v3.21/main
+  packages:
+    - ca-certificates-bundle
+    - tzdata
+    - alpine-baselayout-data
```
waveywaves, line 19: Non-blocking, but I think adding `alpine-release` is still the right shape if we are going to own this Alpine-based image here.

I tested the current config locally with the same apko version pinned in CI (`v1.2.16`), and both local builds and the PR's `build-image (static-base)` logs warn that `/etc/os-release` is missing for all four arches. Adding `alpine-release` removes the warning and adds `/etc/os-release`, `/usr/lib/os-release`, and `/etc/alpine-release`.

I also checked the current replacement image, `ghcr.io/wolfi-dev/static:alpine`; it includes those Alpine identity files. Its per-arch layer size is about 319 KiB. Our current apko output is about 310 KiB per arch without `alpine-release`, and about 321–322 KiB per arch with it. So this would line the self-built image up with the replacement Alpine static image metadata without materially changing the size.

.github/workflows/ci.yaml
```diff
@@ -149,13 +163,43 @@ jobs:
...
+          set -euo pipefail
+          tarball="apko_${APKO_VERSION}_linux_amd64.tar.gz"
+          curl -fsSL -o "${tarball}" \
+            "https://github.com/chainguard-dev/apko/releases/download/v${APKO_VERSION}/${tarball}"
+          echo "${APKO_SHA256}  ${tarball}" | sha256sum -c -
+          tar -xzf "${tarball}"
+          apko="$(pwd)/apko_${APKO_VERSION}_linux_amd64/apko"
+          if [[ "${EVENT_NAME}" == "pull_request" ]]; then
+            # Validate the build on PRs without pushing.
+            "${apko}" build "${CONFIG}" "${IMAGE}:latest" output.tar
```
waveywaves, line 201: Non-blocking follow-up/consistency note, not required for the ownership/base-image direction.

apko's default VCS detection works when it can inspect a normal Git checkout, but it is best-effort. In a full checkout it added source/revision annotations on the OCI index for me; when VCS probing failed, it silently produced only `org.opencontainers.image.created`. Even in the successful case, the per-platform config labels only had `created`, while the existing Dockerfile images explicitly set source/description/license labels.

If we want parity with the Dockerfile images, we could pass explicit `--annotations` for source, license, and description to both `apko build` and `apko publish`. Otherwise I think this can be deferred.

.github/workflows/ci.yaml
```diff
@@ -149,13 +163,43 @@ jobs:
...
+          else
+            "${apko}" publish "${CONFIG}" "${IMAGE}:latest" "${IMAGE}:${GITHUB_SHA}"
```
waveywaves: Non-blocking tag consistency note, not a blocker for this PR.

The Dockerfile path uses `docker/metadata-action`, so pushed images get the branch tag, `sha-*`, and `latest` on the default branch. The apko path publishes exactly these two tags: `latest` and the raw full `${GITHUB_SHA}`.

That works if consumers use `latest` or digests. If we want conventions to be consistent across `tekton/images/*`, we can mirror the Docker metadata-action tag scheme here or track that as a follow-up.

waveywaves (approved):

PR #3519, opened 2026-07-17
fix: clean up GitHub release title and add optional draft param

jkhelil:
# Changes

The `create-draft-release-oci` task (used by e.g. tektoncd/operator's release
pipeline) never set the GitHub release title (`gh release create` was never
given `--title`), while the H1 line in the generated release body fell back
to a synthesized `Release ${VERSION}` name whenever none was given. This
produced a duplicated, quoted title once a maintainer copied that H1 into
the release name by hand, e.g.:

```
Tekton Operator release v0.79.2 "Release v0.79.2"
```

This PR:

- Only appends the quoted release name when one is actually provided,
  instead of synthesizing a redundant fallback.
- Explicitly sets `--title` on `gh release create`, so the release always
  gets a clean, correct title (e.g. `Tekton Operator release v0.79.2`) even
  while left as a draft, instead of sitting untitled until a human edits it.
- Adds an optional `draft` param (default `"true"`, preserving current
  behavior for all existing callers) so projects that don't need manual
  curation for every release — e.g. routine patch releases — can opt to
  publish the release directly instead of leaving an unpublished draft
  behind. When publishing directly, `--latest=false` is passed so GitHub's
  publish-date-based "Latest" auto-marking can't be tripped by an older
  release branch being published after a newer one; callers that do want
  "Latest" set can still do so afterwards via `gh release edit --latest`.

Both the `create-draft-release-oci` Task and the wrapping `release-draft-oci`
Pipeline get the new `draft` param, for symmetry. The older, GCS/`hub`-based
`create-draft-release` task and `release-draft` pipeline are intentionally
left untouched, to keep this change scoped to the actively used OCI variant.

/kind bug

# Submitter Checklist

- [x] Includes docs (README update for the new `draft` param)
- [x] Commit messages follow commit message best practices

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tekton/resources/release/base/github_release_oci.yaml
+++ b/tekton/resources/release/base/github_release_oci.yaml
@@ -37,8 +37,8 @@ spec:
     default: "release"
   - name: draft
     description: |
-      Whether to create the GitHub release as a draft pre-release for manual
-      review (the default, preserving existing behavior). Set to "false" to
+      Whether to create the GitHub release as a draft for manual review
+      (the default, preserving existing behavior). Set to "false" to
       publish the release directly instead.
     default: "true"
   workspaces:
@@ -88,9 +88,12 @@ spec:
         fi
         printf '%s' "${TITLE}" > $HOME/release-title.txt
 
+        # The title above is already set on the release itself via
+        # `gh release create --title`, and GitHub renders it prominently
+        # above the description — repeating it as the first line of the
+        # body here would just be redundant, so the body starts straight
+        # into the content instead.
         cat <<EOF | tee $HOME/release.md
-        ${TITLE}
-
         <!-- For major releases, add a tag line
         # 🎉 [Tag Line - to be done] 🎉
         -->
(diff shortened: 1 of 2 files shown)


mathur07 (approved):
/lgtm

tekton/resources/release/base/github_release_oci.yaml
```diff
@@ -295,17 +317,29 @@ spec:
...
 
+        if [ "${DRAFT}" = "false" ]; then
+          # Publish directly. Explicitly opt out of GitHub's "Latest"
+          # auto-marking (which is based on publish date, not semver), since
+          # patch releases for older branches can be published out of
+          # chronological order; callers that do want this release marked
+          # "Latest" can pass --latest afterwards with `gh release edit`.
+          STATE_FLAGS="--latest=false"
+        else
+          STATE_FLAGS="--draft --prerelease"
```
anithapriyanatarajan: NIT: @jkhelil Could you please reconsider the purpose of these flags. Marking a release as "Draft" is not intended to publish yet, "prerelease" indicates published but unstable. So I would recommend either this be draft or pre-release. I hope the intention is for Patch releases not to be latest?

jkhelil: Yeah right, just copied whatever was there. Fixed now

anithapriyanatarajan: Thank you.

